PR #7, opened 2019-04-11
Join_method_realization_by_Stas_Gostev

StasGostev:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,27 @@
  *
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function() {
-    // write code here
+  [].__proto__.join2 = function(symbol) {
+    let length = this.length;
+    // if recieved empty array , returns empty string
```
roman-mate: It's clear what is happening here and this comment is not very useful. Remove the comment.

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,27 @@
...
+    let result = length
+      ? `${this[0]}`
+      : '';
+
+    // typeof argument is undefined returns join by ','
```
roman-mate: Same here, everything is clear, no need for a comment.

src/arrayMethodJoin.js
StasGostev: Done.

src/arrayMethodJoin.js
StasGostev: Done

PR #11, opened 2019-04-11
solution

VladimirTi:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,23 @@
  *
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function() {
-    // write code here
+  [].__proto__.join2 = function(separator) {
+    let result = '';
+    separator = separator + '';
```
YuriiHoliuk: do not mutate arguments

YuriiHoliuk: after this string separator never be undefined

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,23 @@
  *
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function() {
-    // write code here
+  [].__proto__.join2 = function(inputStr) {
+    let result = '';
+    let separator;
+    inputStr === undefined ? separator = ',' : separator = String(inputStr);
```
YuriiHoliuk: You do not mutate separator - use const
```
    separator = inputStr === undefined ? ',' : String(inputStr);
```

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function() {
-    // write code here
+  [].__proto__.join2 = function(inputStr) {
+    let result = '';
+    const separator = inputStr === undefined ? ',' : String(inputStr);
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
```
YuriiHoliuk: it is always will be false

YuriiHoliuk: I was wrong, it is ok

PR #15, opened 2019-04-11
Solution

GitKurmax:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,22 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function() {
-    // write code here
+  [].__proto__.join2 = function(arg) {
+    return this.reduce((str, item) => {
+      if (arg === undefined) {
+        arg = ',';
+      }
+      if (arg === null) {
+        arg += '';
```
roman-mate: is this the same as 'arg' = ''? Maybe let's change to `arg = ''` for clarity?

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,22 @@
  *
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function() {
-    // write code here
+  [].__proto__.join2 = function(arg) {
+    return this.reduce((str, item) => {
```
roman-mate: using reduce is overkill here. You can just iterate over array elements and concatenate the result.

src/arrayMethodJoin.js
GitKurmax: I didn`t understant your comment. The check (arg === null) needed when the function called with argument null. In this case arg must become string 'null'.

src/arrayMethodJoin.js
```diff
@@ -5,9 +5,24 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function() {
-    // write code here
+  [].__proto__.join2 = function(arg) {
+    let resString = '';
+    if (arg === undefined) {
+      arg = ',';
+    }
+    if (arg === null) {
+      arg += '';
```
roman-mate: This is a very unintuitive way to write that `arg = 'null'.`

src/arrayMethodJoin.js
```diff
@@ -5,9 +5,24 @@
...
+    }
+    for (let item of this) {
+      if (this.indexOf(item) === this.length - 1 || !arg) {
```
roman-mate: indexOf uses loop inside. So for every item you are running a loop. 

If there are N elements in the array, this is doing N^2 operations. 

Rewrite this loop to avoid doing N^2 operations.

PR #16, opened 2019-04-11
solution 

ilovepinkpony1:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,27 @@
  *
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function() {
-    // write code here
+  [].__proto__.join2 = function(token = ',') {
+    let str = '';
+    let addToken = String(token);
```
YuriiHoliuk: addToket - is function name. To avoid using let, you may assign to const with ternary operator

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,27 @@
...
+    if (token === null) {
+      addToken = String(null);
+    }
+    for (let i = 0; i < this.length; i++) {
+      if (
+        this[i] === null
+        || this[i] === undefined // eslint-disable-line
+      ) {
+        str += addToken;
+        continue;
```
YuriiHoliuk: continue is hard to maintain. Rewrite for loop without it and without return. Just use conditions to add necessary parts to string, then return just once after the loop

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,20 @@
...
+    const element = token === null ? String(null) : String(token);
+    for (let i = 0; i < this.length; i++) {
+      if (
+        this[i] === null
+        || this[i] === undefined // eslint-disable-line
+      ) {
+        str += element;
+      } else if (i === this.length - 1) {
+        str += this[i];
+      } else str += this[i] + element;
```
YuriiHoliuk: use brackets for last else to. You may not use them only with single if

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,22 @@
  *
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function() {
-    // write code here
+  [].__proto__.join2 = function(token = ',') {
+    let str = '';
+    const element = token === null ? String(null) : String(token);
```
YuriiHoliuk: Seems like 
```javascript
    const element = String(token);
```
will be the same

PR #17, opened 2019-04-12
added testcase to detect default method usage

ivadimko:


src/arrayMethodJoin.test.js
```diff
@@ -9,6 +9,11 @@ test('join2 is added to [].__proto__', () => {
     .toBeInstanceOf(Function);
 });
 
+test(`join2 doesn't call default join`, () => {
+  expect([].join2.toString().includes('.join'))
+    .toBe(false);
+});
```
mgrinko: it is overkill
I think we believe they won't use it

ivadimko: https://github.com/mate-academy/js_array-method-push/pull/5/files

mgrinko: Ok let's add the test

mgrinko: @ivadimko lets check for `.join(`

PR #18, opened 2019-04-14
Yevhen Kryvdiuk

Yevhen-Kryvdiuk:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,30 @@
  *
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function() {
-    // write code here
+  [].__proto__.join2 = function(str) {
+    let string = '';
+    let separator = str;
+    if (separator === undefined) {
+      separator = ',';
```
YuriiHoliuk: use function default parametres
https://developer.mozilla.org/uk/docs/Web/JavaScript/Reference/Functions/Default_parameters

Then you could initialize const separator without if

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,30 @@
...
+    }
+    if (separator === null) {
+      separator = 'null';
+    }
+    if (this.length === 0) {
+      return string;
+    } else {
```
YuriiHoliuk: you don\`t need `else` after `return`. Moreover, you can miss it because with `length === 0` loop  will work 0 times

src/arrayMethodJoin.js
Yevhen-Kryvdiuk: 1

Yevhen-Kryvdiuk: I have to use let
`let separator = str;
    if (separator === null) {
      separator = 'null';
    }`
for passing the test
`test(`join with null`, () => {
  expect(source.join2(null))
    .toBe('0null1null2null3');
});`

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,23 @@
...
+  [].__proto__.join2 = function(str = ',') {
+    let string = '';
+    let separator = str;
+    if (separator === null) {
+      separator = 'null';
+    }
+    this.forEach((currentValue, index, array) => {
+      const supplemented = currentValue === undefined || currentValue === null
+        ? ''
+        : currentValue;
```
roman-mate: you can replace this ternary operator with 
const supplemented = currentValue || '';

PR #20, opened 2019-04-15
Ostap_Solution

ost-ptk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,10 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    return String(this.length === 0 ? ''
+      : this.map(el => el === undefined || el === null ? '' : el)
+        .reduce((sum, curent) => sum + String(separator) + curent));
```
ivadimko: I'd recommend using the next pattern for splitting ternary to separate lines:
```
condition
  ? ... code if true
  : ... code if false
```

ivadimko: It is not clear enough why do you initialize a string with `String` constructor. `separator` is already `string`. I think there is no need to use `String` object

ivadimko: The variable `sum` in `reduce` is not related to the current case. It's better to use a general definition like `accumulator` or, for the local case, `result`, `string`, `resultString` or smth else. `sum` is more for numbers

vadim-ilchenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,13 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    return String(this.length === 0
+      ? ''
+      : this.map(el => el === undefined || el === null
+        ? ''
+        : el)
+        .reduce((result, curent) => result + String(separator) + curent));
```
ivadimko: suggestion: split code to multiple lines only when line almost reaches max-length (80 chars)
try to solve without `String()` because it's a bit hard to read now

vadim-ilchenko (changes requested):

vadim-ilchenko (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,15 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let separatorStirng = separator;
+    if (typeof separator !== 'string') {
+      separatorStirng = JSON.stringify(separator);
+    }
+    return String(this.length === 0
+      ? '' : this.map(el => el === undefined || el === null
```
danmysak: Не вкладай тернарні оператори.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,15 @@
...
+        ? '' : el)
+        .reduce((result, curent) => result + separatorStirng + curent));
```
danmysak: Краще просто циклом.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,15 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let separatorStirng = separator;
+    if (typeof separator !== 'string') {
+      separatorStirng = JSON.stringify(separator);
```
danmysak: `String(separator)`

PR #21, opened 2019-04-15
solved by @laverka

valeriitovstyk:


src/arrayMethodJoin.js
```diff
@@ -3,10 +3,21 @@
 /**
  * Implement method join
  */
+
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(arg = ',') {
+    let parsedString = '';
```
roman-mate: you are not really parsing anything here. 'res' or 'result' would be a better name.

src/arrayMethodJoin.js
```diff
@@ -3,10 +3,21 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] === 'undefined' || this[i] === null) {
+        parsedString += arg;
+      } else if (i === this.length - 1) {
+        parsedString += this[i];
+      } else {
+        parsedString += '' + this[i] + arg;
```
roman-mate: why do you need an empty string here?

src/arrayMethodJoin.js
```diff
@@ -3,10 +3,21 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(arg = ',') {
+    let parsedString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] === 'undefined' || this[i] === null) {
+        parsedString += arg;
+      } else if (i === this.length - 1) {
```
roman-mate: should it be 'else if' or just 'if' ?

src/arrayMethodJoin.js
valeriitovstyk: rewritten from scratch, not the best  even not good one :/

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,26 @@
...
+
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(arg = ',') {
+    if (arg === null) { arg = 'null'; }
+    let result = '';
+    const arrayLength = this.length;
+    for (let i = 0; i < arrayLength; i++) {
+      if (arrayLength === 0) return result;
```
roman-mate: why do this inside of the loop?

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,26 @@
...
+      if (arrayLength > 0) {
```
roman-mate: is this if needed?

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,26 @@
...
+        if (i === arrayLength - 1) {
+          result += this[i];
+        } else {
+          if (this[i] === null || this[i] === undefined) {
+            this[i] = '';
```
roman-mate: you are modifying an array here.

src/arrayMethodJoin.js
valeriitovstyk: for no reason, my fault, thanks

src/arrayMethodJoin.js
valeriitovstyk: :( done

src/arrayMethodJoin.js
valeriitovstyk: for no reason, my fault, thanks

PR #68, opened 2019-08-22
solution

Rafler:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let str = '';
+    const sep
+      = separator === undefined ? ',' : separator == null ? 'null' : separator;
```
danheim: тернарный оператор не может быть вложен в другой тернарный

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let str = '';
+    let sep = '';
+    separator === undefined
+      ? (sep = ',')
```
irenhh: before writing condition you have to write the element you're working with
`sep = ...`

irenhh: you still have nested ternary operators

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let str = '';
+    let sep = '';
```
irenhh: `sep` is a bad name

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
+    separator === undefined
+      ? (sep = ',')
+      : separator == null
+        ? (sep = 'null')
+        : (sep = separator);
+
+    for (let i = 0; i < this.length; i++) {
+      this[i] === undefined || this[i] === null
+        ? (str += sep)
```
irenhh: same here

irenhh:
why do you still have nested ternary operators?

PR #74, opened 2019-08-25
solution

youvovk:
[LINK]()

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,18 @@
  * Implement method join
  */
 function applyCustomJoin() {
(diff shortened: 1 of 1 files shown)


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,18 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
+    for (let i = 0; i < this.length; i++) {
```
irenhh: new line before and after for loop

PR #76, opened 2019-08-25
fs_jul19 Alex Andrutskii solution

aleks-and:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+
+      if (currentElement === null
+        || currentElement === undefined
+        || currentElement === '') {
+        currentElement = '';
+      }
+
+      i !== this.length - 1
+        ? stringFromArray += currentElement + separatorForJoin
+        : stringFromArray += currentElement;
```
irenhh: specify the element you're working with firstly
`stringFromArray = ...`

irenhh (changes requested):

src/arrayMethodJoin.js
aleks-and: Fixed? If I understand correctly.

irenhh (approved):

PR #79, opened 2019-08-25
long again but working

Volodsher:
and, of course: Muchas gracias!!!

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,34 +4,22 @@
  * Implement method join
  */
 
-// let x = [0];
-// function joining(separator) {
-//   let str = '';
-//   let middleString = '';
-//   if (separator !== undefined) {
-//     str += x[0];
-//     for (let a = 1; a < x.length -  1; a++) {
-//       middleString = middleString + separator + x[a];
-//     };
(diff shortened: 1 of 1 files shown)


src/arrayMethodJoin.js
```diff
@@ -3,9 +3,37 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
+    }
+    let str = '';
+    let middleString = '';
```
irenhh: `str` and `middleString` are not really good names, I don't see the difference between them from their names

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,37 @@
...
+    if (this.length === 0) {
+      return str;
+    };
+    if (this.length === 1) {
+      str += this[0];
+      return str;
+    };
+    if (separator !== undefined) {
+      str += this[0];
+      for (let a = 1; a < this.length - 1; a++) {
```
irenhh: why `a`?

PR #81, opened 2019-08-25
'Solution'

Drop777:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
+    if (separator === null) {
+      newSeparator = 'null';
+    }
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1) {
+        newString += this[i];
+        break;
+      }
+      this[i] === null || this[i] === undefined ? newString += ','
+        : newString += this[i] + newSeparator;
```
irenhh:
```
newString += this[i] === null || this[i] === undefined 
        ? ','
        : this[i] + newSeparator;
```

irenhh: first of all you need to write the element that you defining and then condition

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
+    if (separator === undefined) {
+      newSeparator = ',';
+    }
+    if (separator === null) {
+      newSeparator = 'null';
+    }
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1) {
+        newString += this[i];
+        break;
```
irenhh: you can return result immediately

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
+      }
+      this[i] === null || this[i] === undefined ? newString += ','
+        : newString += this[i] + newSeparator;
+    }
+    return newString;
```
irenhh: new line before return

PR #83, opened 2019-08-26
create doc

aleks-mos:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,18 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
     // write code here
+    let stringJoin = '';
+    const joinSeparator = (separator === undefined) ? ',' : separator;
+    for (let i = 0; i < this.length; i++) {
```
irenhh: new line before and after for loop

irenhh: and after as well

PR #90, opened 2019-10-02
array method join

Volodymyr-Roiuk:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let resultStr = '';
+
+    if (separator === undefined) {
+      separator = ',';
+    }
+
+    separator = '' + separator;
```
earthhuman41: к строке можно привести так separator.toString()

Volodymyr-Roiuk: Done

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
gestap: use default value
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
-    // write code here
+    let resultStr = '';
+
+    if (separator === undefined) {
+      separator = ',';
+    }
+
+    separator = separator === null ? 'null' : separator.toString();
```
gestap: create a variable with result of `String(separator)`

gestap: then use it

src/arrayMethodJoin.js
Volodymyr-Roiuk: it was really good idia

src/arrayMethodJoin.js
Volodymyr-Roiuk: Done

PR #104, opened 2019-10-04
Solution

frucelestine:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,27 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let sep = separator;
+    let result = '';
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
+      if (sep === undefined) {
```
gestap: you can add a default value to separator and remove this `if`

`separator  = ','`

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,27 @@
...
+        sep = ',';
+      }
+      if (sep === null) {
+        sep = 'null';
```
gestap: you can use `String(separator)` and remove this if

PR #108, opened 2019-10-04
Solution

vitya63:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,11 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

PR #109, opened 2019-10-04
solution

lubremets:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,18 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+    const arrayLength = this.length;
+    for (let i = 0; i < arrayLength; i++) {
+      if ((this[i] !== null) && (this[i] !== undefined)) {
+        str += this[i];
+      }
+      if ((i < arrayLength - 1)) {
```
gestap:
```suggestion
      if (i < arrayLength - 1) {
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,18 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+    const arrayLength = this.length;
+    for (let i = 0; i < arrayLength; i++) {
+      if ((this[i] !== null) && (this[i] !== undefined)) {
```
gestap:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
```

PR #111, opened 2019-10-07
'Solution'

Xandr1:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,8 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

PR #112, opened 2019-10-07
Solution

FeelineKing:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
...
-    // write code here
+    let joinedString = '';
+    let defaultSeparator;
+
+    if (separator === undefined) {
+      defaultSeparator = ',';
+    } else {
+      defaultSeparator = '' + separator;
```
gestap: better to write like this `String(separator)`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  Array.prototype.join2 = function(separator = ',') {
+    let joinedString = '';
+
+    if (separator !== undefined) {
```
gestap: now no need to do this check, you already add default value.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+      separator = String(separator);
```
gestap: better to create a variable and use it
```suggestion
      const separatorStr = String(separator);
```

PR #115, opened 2019-10-07
'solution_vitalii'

VitaliyStarodub:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,27 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
```
gestap: remove
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,27 @@
...
+    const checkedSeparator = separatorCheck(separator);
```
gestap: you can use default method `String(separator)`

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,27 @@
...
+    let string = '';
+    for (let i = 0; i < this.length; i++) {
+      const checkedElement = arrayElementCheck(this[i]);
+      if (i < this.length - 1) {
+        string += checkedElement + checkedSeparator;
+      } else {
+        string += checkedElement;
+      }
+    }
+    return string;
```
gestap:
```suggestion

    return string;
```

PR #150, opened 2020-01-24
AnastasiiaBardyuzha:  js_array-method-join solution

AnastasiiaBardyuzha:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,37 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let joinSeparator = separator;
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    if (joinSeparator === undefined) {
```
irenhh: add default case to arguments

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,37 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let joinSeparator = separator;
```
irenhh: there is no need for this

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+    }
+
+    let string = String(this[0]);
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
+
+      const point = String(separator) + this[i];
```
irenhh: use interpolation

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+    }
+
+    let string = String(this[0]);
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
+
+      const point = `${String(separator)}${this[i]}`;
```
irenhh: if you use interpolation you don't need String() any more

PR #168, opened 2020-01-27
Bohdan Prymyak join

Arimazz:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+    for (let i = 0; i < arr.length; i++) {
+      if (String(separator) === 'null') {
+        if (i < arr.length - 1) {
+          joined += arr[i] + 'null';
+        } else {
+          joined += arr[i];
+        }
+      } else if (arr[i] === null || arr[i] === undefined) {
+        joined += '' + separator;
+      } else if (i < arr.length - 1) {
```
irenhh: you check this condition twice, think about how to reduce the amount of `if`s

PR #173, opened 2020-01-27
array-method-join

serverieva:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let outStr = '';
```
irenhh: outStr  is too abstract name

PR #177, opened 2020-01-28
solved

LeonidShv:
https://github.com/LeonidShv/js_array-method-join/blob/develop/src/arrayMethodJoin.js

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,25 @@
  * Implement method join
  */
 function applyCustomJoin() {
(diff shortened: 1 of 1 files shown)


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+      if (this[i] === null || this[i] === undefined) {
+        result += separator;
+        continue;
+      }
+
+      result += this[i];
+
+      if (i === this.length - 1) {
+        break;
+      }
```
irenhh: why do you add this manually, this is the condition in the for loop

PR #233, opened 2020-05-26
solution

aleksey-10:
:3

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -5,7 +5,24 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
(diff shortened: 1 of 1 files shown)


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
...
+    const checkItem = item => (item === null || item === undefined) ? '' : item;
+
+    for (let i = 0; i < this.length; i++) {
+      if (i === 0) {
+        string = `${checkItem(this[i])}`;
+        continue;
+      }
+
+      string += `${separator
+        !== undefined ? separator : ','}${checkItem(this[i])}`;
```
solaryasha, line 22: You could add default separator
`function(separator=',')`

PR #278, opened 2020-08-05
solution

masyatya:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+    }
+
+    let result = '';
+    let separatorCopy = '';
+
+    if (separator === null) {
+      separatorCopy += 'null';
+    } else {
+      separatorCopy += separator;
+    }
```
krajnova, line 19: Are you sure that you need a separatorCopy?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+
+    let result = '';
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        result += this[i];
+      } else {
+        result += '';
+      }
```
vsm2k17, line 19: why do you need this "else" block?

PR #284, opened 2020-08-05
solution

furude-rika:


src/arrayMethodJoin.js
```diff
@@ -7,15 +7,15 @@ function applyCustomJoin() {
...
       }
 
       if (separator === null) {
-        result += 'null' + key;
+        result += 'null' + elem;
       } else {
-        result += separator + key;
+        result += separator + elem;
       }
     }
```
rmnkk, line 21: I'm not sure that we are allowed to use array methods in this task.

furude-rika: I used it as a string method, but I'll remove it if you like

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let elem of this) {
+      if (elem === null || elem === undefined) {
+        elem = '';
+      }
+
+      if (separator === null) {
```
vsm2k17, line 15: no need to do this, it works the same without extra code

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+        result += 'null' + elem;
+      } else {
+        result += separator + elem;
+      }
+    }
+
+    return result.slice(String(separator).length, result.length);
```
vsm2k17, line 22: instead of using slice, you can just add separator if element index not '0' (in this case you need to use 
"for (let i=0; i < this.length; i+=1 )" )

furude-rika: thank you, I'll fix it

PR #297, opened 2020-08-06
Solution

kostyasabada:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,26 @@
...
+
+    let str = '';
+
+    for (let i = 0; i < this.length - 1; ++i) {
+      if ((this[i]) === null || (this[i]) === undefined) {
+        str += separator;
+        continue;
+      }
+
+      str += '' + this[i] + separator;
```
vkryvytskyy, line 20:
```suggestion
      str += this[i] + separator;
```
do you need an empty string here?

kostyasabada: > do you need an empty string here?

Yes, without it doesnt work 100%

vkryvytskyy: then you do something wrong. You could remove it.

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,27 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length < 1) {
+      return '';
+    }
+
+    let str = '';
+
+    for (let i = 0; i < this.length - 1; ++i) {
+      if ((this[i]) === null || (this[i]) === undefined) {
```
Denchepornyuk, line 15:
```suggestion
      if (this[i] === null || this[i] === undefined) {
```

PR #301, opened 2020-08-06
joint

SobchukSasha:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null && i === 0) {
```
vkryvytskyy, line 11: you don't need separate if for first element === null, your second if handle this case too

PR #302, opened 2020-08-06
solution

YuriiRysyn:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,30 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (separator === undefined) {
+        if (i === this.length - 1) {
+          finalStr += this[i];
+          break;
+        }
+        finalStr += this[i] + ',';
+        continue;
+      }
```
vkryvytskyy, line 19: just add default value to separator in 7th line

PR #303, opened 2020-08-06
Solution

mpone:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
```
vkryvytskyy, line 12: method join should not change initial array

PR #308, opened 2020-08-06
Solution

artsashko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1) {
+        string += this[i];
+        break;
+      }
+
+      if (this[i] == null || typeof this[i] === 'undefined') {
+        string += separator;
+      } else {
+        string += this[i] + String(separator);
```
vkryvytskyy, line 19:
```suggestion
        string += this[i] + separator;
```
i think separator is already a string

artsashko: if I`ll remove this, it won`t be work with null as a separator

PR #310, opened 2020-08-06
solution

karinaKudatska:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        str += this[i];
+      } else {
```
Elena-brv, line 13: you don`t need block 'else' here, because the empty string will not visible at the result so it's not necessary to be added.

karinaKudatska: дякую, виправила))

PR #325, opened 2020-08-11
Method join implementation

asbh25:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
```
vkryvytskyy, line 12: initial array should not be modified

asbh25: fixed

PR #403, opened 2020-12-14
solution

Lara-l-art:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] == null || this[i] === undefined) {
+        string += sepp;
+      } else {
+        string += this[i] + sepp;
+      }
+    }
+
+    return string + this[this.length - 1];
```
VinnieJ-2k20, line 27:
```suggestion
    if (this[this.length - 1] == null || this[this.length - 1] === undefined) {
      return string;
    } else {
      return string + this[this.length - 1];
    }
```

VinnieJ-2k20 (changes requested):

VinnieJ-2k20 (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+    if (this.length === 0) {
+      return '';
+    }
+
+    if (separator == null) {
+      sepp = 'null';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1) {
```
aionova, line 20: Make one check for null and undefined in the beginning of your loop. Don't repeat yourself.

aionova (changes requested):

aionova (approved):
Good job!

PR #410, opened 2020-12-15
Solution

olesiatk:


MarchD (approved):
good solution!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+    if (Array.isArray(this) && this.length > 1) {
+      if (this[0] === null || this[0] === undefined) {
+        this[0] = '';
+      }
+
+      let str = `${this[0]}`;
+
+      for (let i = 1; i < this.length; i++) {
+        if (this[i] === null || this[i] === undefined) {
+          this[i] = '';
```
aionova, line 17: Don't mutate the initial array

aionova (changes requested):

src/arrayMethodJoin.js
olesiatk: Fixed

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (Array.isArray(this) && this.length > 1) {
+      let str;
+
+      if (this[0] === null || this[0] === undefined) {
```
aionova, line 11: No need to duplicate this check. You can check this element in a loop too, and make str = '' initially.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (Array.isArray(this) && this.length > 1) {
+      let str;
```
aionova, line 9:
```suggestion
      let str = '';
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
...
+        str = `${this[0]}`;
+      }
+
+      let arrValue;
+
+      for (let i = 1; i < this.length; i++) {
+        if (this[i] === null || this[i] === undefined) {
+          arrValue = '';
+        } else {
+          arrValue = `${this[i]}`;
```
aionova, line 23: When you add element to a string, it becomes string too. No need to use interpolation here.

aionova (changes requested):

src/arrayMethodJoin.js
olesiatk: fixed

src/arrayMethodJoin.js
olesiatk: I made "str" as a string when added its first element as a string.

aionova: Yes, and I wrote that you don't need to duplicate this condition. You can check this element in a loop too, and make str = '' initially.

aionova (approved):

PR #412, opened 2020-12-15
added solution

MarchD:


olesiatk (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
+
+    let joinedElements = '';
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        this[i] = '';
```
aionova, line 16: You mutate the original array here and you shouldn't

aionova (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const array = [...this];
```
aionova, line 8: No need to include data type in the name of a variable. We know that it's an array. 
```suggestion
    const elements = [...this];
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+      return '';
+    }
+
+    let joinedElements = '';
+
+    for (let i = 0; i < array.length - 1; i++) {
+      if (array[i] === undefined || array[i] === null) {
+        array[i] = '';
+      }
+      joinedElements = joinedElements + array[i] + separator;
```
aionova, line 20:
```suggestion
      joinedElements += array[i] + separator;
```

aionova (approved):
Good job!

PR #415, opened 2020-12-15
Solution

khrulenko:


dima-bond (changes requested):
This task can be managed with 1 loop and less nested ifs. You should check all parameters first, and then use them in a loop.

dima-bond (approved):
gg wp!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+      return '';
+    }
+
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        str += separator;
+      } else {
+        str += `${this[i]}`;
```
aionova, line 18: When you add element to a string, it becomes string too. No need to use interpolation here.

aionova (approved):
Nice!

PR #428, opened 2020-12-15
solution

chri911:


HaidaiAndrii (approved):
Good job)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+    let connector = separator;
+
+    if (separator === null) {
+      connector = 'null';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (i !== this.length - 1 && this.length !== 0) {
```
aionova, line 16:
```suggestion
      if (i !== this.length - 1) {
```
If this.length = 0, you can't iterate through it.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+        if (this[i] === undefined || this[i] === null) {
+          str += connector;
+        } else {
+          str += this[i] + connector;
+        }
+      } else {
```
aionova, line 22: This condition works if this.length = 0. Then how can you write str += this[i], if this[i] doesn't exist?

aionova (changes requested):

src/arrayMethodJoin.js
chri911: Fixed

aionova (approved):
good job!

PR #435, opened 2020-12-16
solution

Droshik:


Alexandr-Petrenko (approved):
Хорошая работа!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+      return '';
+    }
+
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] === 'undefined' || this[i] === null) {
+        result += separator;
+        continue;
+      }
```
aionova, line 18:
```suggestion
      if (this[i] === 'undefined' || this[i] === null) {
        result += separator;
      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+
+      if (i !== this.length - 1) {
+        result += this[i].toString() + separator;
```
aionova, line 21:
```suggestion
        result += this[i] + separator;
```
If you add element to a string, it becomes string

aionova (approved):

PR #436, opened 2020-12-16
solution

YuriiFitsyk:


Iryna713 (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (i !== this.length - 1) {
+        if (this[i] === null || this[i] === undefined) {
+          result += separator;
+        } else {
+          result += this[i];
+          result += separator;
+        }
+      } else {
+        result += this[i];
```
aionova, line 19: You need to check all the elements for null and undefined. Now you don't check the last one

aionova (changes requested):

aionova (approved):
good job!

PR #443, opened 2020-12-17
Solution

Dima-Semenov:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
```
MarchD, line 11: If you change the condition, your code will be shorter
For example : `this[i] !== null ... and so on`

MarchD (approved):
Good job:)

Denchepornyuk (approved):

PR #468, opened 2021-02-14
solution

Ostapiuss:
solution of array method join

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -5,7 +5,36 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let string = '';
+
+    if (typeof (separator) === 'undefined') {
+      for (let i = 0; i < this.length; i++) {
+        this[i] = this[i].toString();
+
+        if (i === this.length - 1) {
+          string += this[i];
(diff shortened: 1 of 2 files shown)


taraseman (approved):
good

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,36 @@
...
+
+        if (i === this.length - 1) {
+          string += this[i];
+        } else {
+          string += this[i] + ',';
+        }
+      }
+    }
+
+    if (typeof (separator) !== 'undefined') {
```
Vladimir-Beliy, line 22:
```suggestion
    else {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,36 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let string = '';
+
+    if (typeof (separator) === 'undefined') {
```
Vladimir-Beliy, line 10:
```suggestion
    if (typeof separator === 'undefined') {
```

Vladimir-Beliy (changes requested):

Vladimir-Beliy (approved):

PR #496, opened 2021-03-03
Solution

danylo-onishchuk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+    let result = '';
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    if ((this[0] === null) || (this[0] === undefined)) {
+    } else {
+      result = `${this[0]}`;
+    }
```
danheim, line 17: looks weird, why do you need this if?

danheim (changes requested):

Denchepornyuk (approved):

PR #539, opened 2021-04-21
Solution

Burunduk003:


IvanGrekov (approved):
LGTM

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let separatorString = separator;
```
mykola-yarchuk, line 9: redundant

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+
+    if (separator === null) {
+      separatorString = 'null';
```
mykola-yarchuk, line 12: that will work as you want anyway if you try to `result += separatorString`, so you can remove this condition

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+    } else {
+      separatorString = '' + separator;
+    }
```
mykola-yarchuk, line 15: redundant in all cases

mykola-yarchuk (changes requested):

mykola-yarchuk (approved):
great

PR #545, opened 2021-04-24
added custom join method

Ihor-Karpyn:


den-bulaev:
Nice work!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const arr = [...this];
```
mykola-yarchuk, line 8: redundant, you can avoid using an extra array, just write what you need into resultStr, without saving a copy of the array

mykola-yarchuk (changes requested):
try to avoid extra array copy

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const arr = this;
```
mykola-yarchuk, line 8: why don't just use this instead of creating arr?

mykola-yarchuk (changes requested):

src/arrayMethodJoin.js
Ihor-Karpyn: I think it's easier to understand

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultStr = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        resultStr += '';
+      } else {
+        resultStr += this[i];
+      }
```
mykola-yarchuk, line 15:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        resultStr += this[i];
      }
```

mykola-yarchuk (approved):

PR #555, opened 2021-04-26
Solution

iryna-vys:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const arr = [...this];
+
+    for (let i = 0; i < arr.length; i++) {
+      if (arr[i] === null || arr[i] === undefined) {
+        arr[i] = '';
```
igor-stasiv, line 13: Here gonna be better if you add at least a separator to the final string, for example, if you have [undefined, 123, null, 'abc'] result will be like ', 123, , abc'. It indicates, that there was a value in the array. And it is how method .join really work

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+      }
+
+      if (i !== arr.length - 1) {
+        result = result + arr[i] + separator;
```
igor-stasiv, line 17:
```suggestion
      if (i !== arr.length - 1) {
        result += arr[i] + separator;
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const arr = [...this];
```
igor-stasiv, line 9: Don't create a copy of the input array, you just can use .this

igor-stasiv (changes requested):

igor-stasiv (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
varseniuk, line 15:
```suggestion
      if (this[i] !== null || this[i] !== undefined) {
        result += this[i];
      }
```
You can use only 1 check instead of 2. If you will check that your element is NOT null or NOT undefined then you add this element to the string. If element is null/undefined - you don't add nothing

varseniuk (approved):
Good! 👍 
But pay attention to my comment and see how you can make your code more simple

PR #556, opened 2021-04-26
Join Method

leonid-vegera:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
```
Namart, line 8: if loop will not start (length === 0), function return 'string' anyway
this condition is redundant

Namart (approved):
good job)

src/arrayMethodJoin.js
leonid-vegera: Agree

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
```
```suggestion
    let concatenatedString = '';
```
Bad idea to name variables with the data-type name. It's better to use `joinedString` or `concatenatedString` (or even `result` at least)

varseniuk (approved):
Great solution! 👍 
Pay attention to the namings

PR #557, opened 2021-04-26
Solution-001

senkiv-oleh:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const arr = [...this];
+
+    let joiner = '';
+
+    for (let i = 0; i < arr.length; i++) {
+      if (arr[i] == null) {
```
Namart, line 13: bad practice too use non strict comparison, better use strict one

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const arr = [...this];
```
Namart, line 8: no need to spread 'this'. you can simply assing it to variable

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+
+    let joiner = '';
+
+    for (let i = 0; i < arr.length; i++) {
+      if (arr[i] == null) {
+        joiner += '' + separator;
+        continue;
```
Namart, line 15: a bit tricky construction with continue

Namart (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
```
```suggestion
    let concatenatedString = '';
```
'str' is a really bad name for the variable. Use smth more meaningful

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+      if (this[i] === null || this[i] === undefined) {
+        str += separator;
+        continue;
+      }
+
+      if (i !== this.length - 1) {
+        str += this[i] + '' + separator;
+      } else {
+        str += this[i];
+      }
```
varseniuk, line 20:
```suggestion
      if (this[i] !== null || this[i] !== undefined) {
        str += this[i];
      }

      if (i < this.length - 1) {
        str += separator;
      }
```

Look, here you can make your code work faster, using 2 checks instead of 3 (and not using `continue`).
First - you check if your element is not `null` or `undefined`. If so - you add this element to the string. Then you check if it's not the last element (i < this.length - 1), and if so - add separator.
If your element is `null` or `undefined` - you just add separator (if it's not the last element in array)

varseniuk (approved):
Nice, but you can make it even better.
Please pay attention to my comment and try to make your cod more simple.

PR #563, opened 2021-04-26
Solution

progmari:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+
+    this.forEach((e, i) => {
```
Andrey203, line 10: На мой взгляд название переменных можно сделать более понятным. e ---> elem    i ---> index

Andrey203 (approved):
Выглядит круто!

src/arrayMethodJoin.js
varseniuk, line 10:
```suggestion
    this.forEach((e, i) => {
```
Let's try not to use another array methods while implementing this one. Use simple `for` loop here

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
```
```suggestion
    let concatenatedString = '';
```
`str` is a really bad name for the variable. Let's use smth more meaningful

```suggestion
    let concatenatedString = '';
```
`str` is a really bad name for the variable. Let's use smth more meaningful

varseniuk (changes requested):
Your logic is correct and the solution is good, but let's not use array methods while implementing other ones.
Change your `forEach` method to a simple `for` loop and it's done.

varseniuk (approved):
Great 🚀

PR #565, opened 2021-04-26
develop

dmytro-abram:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    const array = this;
```
chumasov, line 9: I think You don't need to copy the array. Just use 'this' everywhere.

chumasov (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+    let result = '';
+
+    for (let i = 0; i < array.length; i++) {
+      if (array[i] === null || array[i] === undefined) {
+        result += '';
+      } else {
+        result += array[i];
+      }
```
ArtemTopolskyi, line 17:
```suggestion
      if (array[i] !== null && array[i] !== undefined) {
        result += array[i];
      }
```

ArtemTopolskyi (approved):

ArtemTopolskyi (changes requested):
fix

ArtemTopolskyi (approved):

PR #569, opened 2021-04-26
solution

dmitry-puhliakov:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+      for (let element of this) {
+        if (element === undefined || element === null) {
+          element = '';
+        }
+
+        joinString += `${element}${separator}`;
+      }
+    }
+
+    return joinString.slice(0, joinString.lastIndexOf(separator));
```
varseniuk, line 21: Really clever solution 👍 , but let's try in these tasks not to use another array methods.

varseniuk (changes requested):
Nice solution, but let's get rid of `slice` method.
You should not use common array methods while implementing another ones

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    const len = this.length;
+    let joinString = '';
+
+    if (len !== 0) {
```
```suggestion
    if (len !== 0) {
```
This check is redundant. If `this.length === 0` then loop will not even start:
'for (let i = 0; i < 0; i++)'
0 will never be less than 0 so the loop will not have even one iteration

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+        let element = this[i];
+
+        if (element === undefined || element === null) {
+          element = '';
+        }
+
+        if (i !== len - 1) {
+          joinString += `${element}${separator}`;
+        } else {
+          joinString += `${element}`;
```
varseniuk, line 23:
```suggestion
        if (element !== undefined || element !== null) {
          joinString += element;
        }

        if (i !== len - 1) {
          joinString += separator;
        }
```
Watch here how your code may be more simple and shorter. In such case you will have only 2 checks instead of  3.
First, you check if `this[i]` is not null/undefined, and if so - add the element to string. If it's null or undefined - will happen nothing.
Then you check if element is no the last one in the array, and if so - you add the separator. If not - do nothing

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+    if (len !== 0) {
+      for (let i = 0; i < len; i++) {
+        let element = this[i];
+
+        if (element === undefined || element === null) {
+          element = '';
+        }
+
+        if (i !== len - 1) {
+          joinString += `${element}${separator}`;
```
varseniuk, line 21:
```suggestion
          joinString += element +separator;
```

varseniuk (approved):
Good! 👍 
Left some comments. Pay attention please

PR #570, opened 2021-04-26
solution

provolovskyi-illia:


ivan-shpynda (approved):
good!

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,25 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null || this[i] === ' ') {
+        resultJoin += separator;
+      } else if (i === this.length - 1) {
+        resultJoin += this[i];
+      } else {
+        resultJoin += this[i] + `${separator}`;
+      }
+    }
```
varseniuk, line 22:
```suggestion
    for (let i = 0; i < this.length; i++) {
      if (this[i] !== undefined || this[i] !== null || this[i] !== ' ') {
        resultJoin += this[i];
      }
      if (i !== this.length - 1) {
        resultJoin += separator
      }
    }
```
You can simplify your code and make 2 checks instead of 3.
In the case I've described, you will always add `this[i]` to the new string when `this[i] !==null/undefined` (when this[i] is null or undefined the check won't pass, and will happen nothing). Also, you always adding separator while `i !== this.length - 1`

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,25 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultJoin = '';
+
+    if (this.length === 0) {
+      return resultJoin;
+    }
```
```suggestion
```

It's redundant. Because if `this.length === 0`, then loop will not even start:
`for (let i = 0; i < this.length; i++)` will mean `for (let i = 0; i < 0; i++)`
In this case 0 is never less than 0 and no iterations wil be made

varseniuk (approved):
Nice work! But pay attention to my comments and see how you can simplify your code

PR #571, opened 2021-04-26
Solution

avgusta92:


ivan-shpynda (approved):
Well done!

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,21 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (i !== 0) {
+        str += separator;
+      };
+
+      if (this[i] !== undefined && this[i] !== null) {
+        str += this[i];
+      };
+    };
```
```suggestion
    for (let i = 0; i < this.length; i++) {
      if (this[i] !== undefined && this[i] !== null) {
        str += this[i];
      };
      
      if (i < this.length -1) {
        str += separator;
      };
    };
```
In such case you will always add separator at the end of each iteration while it's not the last element

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
```
```suggestion
    let concatenatedString = '';
```
`str` is a really bad name for the variable. Let's use smth more meaningful

varseniuk (approved):
Good job! 
But pls pay attention to my comments

PR #572, opened 2021-04-26
solution

RuslanaPodshyvalova:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,45 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
varseniuk, line 7:
```suggestion
  [].__proto__.join2 = function(separator) {
```
let's add a default value to the `separator` argument (then there will be no need in your `checkSeparator` function

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,45 @@
...
-    // write code here
+    let strFromArr = '';
+    const separatorSign = checkSeparator(separator);
+
+    if (this.length === 0) {
+      return strFromArr;
+    }
```
varseniuk, line 13:
```suggestion
```
It's redundant. If `this.length === 0` then the loop will not even start:
`for (let i = 0; i < this.length; i++)` will be `for (let i = 0; i < 0; i++)`. Here 0 is never less than 0, so nothing will happen

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,45 @@
...
+      } else {
+        strFromArr += item;
+      }
+    }
+
+    return strFromArr;
   };
 }
 
+function checkSeparator(separator) {
```
varseniuk, line 29:
```suggestion
function checkSeparator(separator) {
```
No need in this function if you will just specify the default value for your `separator` argument

varseniuk (changes requested):
Nice solution! 🔥 
But let's make it even better - there are redundant function in your code that may be simply replaced by giving a default value to the `separator` argument

varseniuk (approved):
Good job! 👍

PR #578, opened 2021-04-27
solution

IvanKolodiazhnyi:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
```
Namart, line 9: you can return 'this' or declare 'newString' earlier and return it. like an option)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+
+    let newString = '';
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        newString += separator;
+        continue;
+      }
+      newString += this[i];
+      newString += separator;
```
Namart, line 20: newString += `${this[i]}${separator}`

Namart (changes requested):

src/arrayMethodJoin.js
IvanKolodiazhnyi: Тут я провтикав)

mykola-yarchuk (approved):
good job but there is a better approach, you can add this[i] to your string if this[i] not undefined and not null. after that, you can add separator if it's not the last element in the array.
In that case, you can avoid adding the last element outside of the loop and if block in line 10.

PR #584, opened 2021-04-27
Solution

SergeyZub:


Grygoriy-Shytikov (approved):
Good job &#128077; 
--

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (this.length === 0) {
+      return '';
+    }
```
```suggestion
```
This check is redundant. If `this.length === 0` the loop won't even start:
`for (let i = 0; i < this.length - 1; i++)` ===> `for (let i = 0; i < 0 - 1; i++)`
here 0 will never be less than -1, so there will be no single iteration in this loop

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
varseniuk, line 19:
```suggestion
      if (this[i] !== null || this[i] !== undefined) {
        result += this[i];
      }
```
You can swap your expressions in `if-else` by places and change the check from `===` to `!==`. In this case, you will be checking if the element is not null or undefined, and if so - add it to the string. If it's null or undefined - nothing will happen in that case.
And this wil help you to get rid of unnecessary `else` statement

varseniuk (approved):
Nice work! 👍 
But pay attention to my comment and see how it may be improved

PR #585, opened 2021-04-27
Solution 1

Grygoriy-Shytikov:


amithos (approved):
Well done)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let strJoin = '';
+
+    if (this.length === 0) {
+      return strJoin;
+    }
```
```suggestion
```
No need for this check. If there are no elements in the array, nothing will happen. Your loop will not even start, and in the end you will still return an empty string either way

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
...
+
+    for (let value of this) {
+      if (this.length === 1) {
+        strJoin += value;
+
+        return strJoin;
+      }
```
varseniuk, line 19:
```suggestion
```

No need for this check. In your code below you always add value to your string (if the length of the array is 1 or more)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
...
+
+      if (value === undefined || value === null) {
```
varseniuk, line 21:
```suggestion
      if (value !== undefined || value !== null) {
```
Let's refactor this code a bit and make it more simple and short. you are repeating 
strJoin += `${separator}${value}`; two times and it's not good.
You can check if the element (`value`) is not null or undefined, and if so - add the value to the string (`strJoin += value`). If it's null or undefined - do nothing.
Also, you need to add a separator (`strJoin += separator`) to your string if it's not the last element of the array

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
...
+        value = '';
+        strJoin += `${separator}${value}`;
+      } else {
+        strJoin += `${separator}${value}`;
+      }
+    }
+
+    const strJoinFix = strJoin.slice(`${separator}`.length, strJoin.length);
```
varseniuk, line 29:
```suggestion
    const strJoinFix = strJoin.slice(`${separator}`.length, strJoin.length);
```
Let's not use in this task built-in array methods (`slice` in this case). Try to make it with base loops and `if`s.
Instead of this `slice` here, you can check in your loop if it's not the last element in the array (for these purposes better to use common `for` loop instead of `for-of`:
`if (i < this.length -1)`
And add a separator to your string, If it's the last element - do nothing

varseniuk (changes requested):
Nice work, but let's make your code a litle better

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let strJoin = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
```
```suggestion
        this[i] = '';
```
Here you are changing the incoming array. The original method `join` does not change the initial array. It should only return a string but the array should be untouched. When you write `this[i] - ''` it means that you make changes to the array (which is not allowed)

varseniuk (changes requested):
You cannot mutate the initial array

varseniuk (approved):
Good 👍

PR #590, opened 2021-04-28
solution

GoldenSpade:


leonid-vegera (approved):
Good job!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    if (this.length === 0) {
+      return string;
+    }
```
```suggestion
```
This check is redundant. If this.length === 0 then the loop will not even start and in the end, the function will still return `string`.
'string`, by the way, is a bad name for the variable. Use something more meaningful in the future

varseniuk (approved):
Good job! 👍 
Left one comment

src/arrayMethodJoin.js
BilliZane: Ok

PR #592, opened 2021-04-28
Solution

olha-tsarova:


ViPiven (approved):
Nice solution!!!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += `${this[i]}`;
+      }
```
varseniuk, line 15:
```suggestion
      if (this[i] !== null || this[i] !== undefined) {
        result += `${this[i]}`;
      };
```
In this case, when `this[i] === null/undefined` we just do nothing, otherwise - add this[i] to the string

varseniuk (approved):
Good job! 🚀 
Pay attention to my comment on how it may be simplified a bit

PR #688, opened 2021-08-03
Solution

alex-glotov67:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let joinedArray = '';
+
+    if (this[0] === null || this[0] === undefined) {
+      joinedArray = '';
+    } else {
+      joinedArray += this[0];
+    }
```
Anna-Tkachenko, line 15: You can also write this code:
```suggestion
    if (this[0] !== null && this[0] !== undefined) {
      joinedArray += this[0];
    }
```

Anna-Tkachenko (approved):
Good job👍

src/arrayMethodJoin.js
varseniuk: Let's get rid of this `if` at all

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+
+    for (let i = 1; i < this.length; i++) {
```
varseniuk, line 17:
```suggestion
    for (let i = 0; i < this.length; i++) {
```
Let's do everything in one loop. Start the loop from the 0 index and check if it's not null and undefined. If so - add `this[i]`, also, add check if it's not the last element. If so - add `separator`

varseniuk (changes requested):
Let's make tour solution a bit shorter

varseniuk (approved):

PR #689, opened 2021-08-04
solution

SashaRudenko:


alex-glotov67 (approved):
Nice solution!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let string = '';
+
+    for (let value = 0; value < this.length; value++) {
```
Ihor-Karpyn, line 11:
```suggestion
    for (let i = 0; value < this.length; value++) {
    const value = this[i];
```
but `for of`  will better here

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+      if (this[value] == null) {
+        string = string + separator;
```
Ihor-Karpyn, line 13:
```suggestion
        string += separator;
```

Ihor-Karpyn (approved):
Great 👌, pay attention to my comments

PR #694, opened 2021-08-04
The task has been done

Moroz-Dmytro:


ihor-jpeg (approved):
Nice code. I like that you used 'continue' instead of 'else'. Good work!

Ihor-Karpyn (approved):
LGTM 🔥

PR #697, opened 2021-08-05
Solution

denyschemeriakov:


ivanchenko-kirill (approved):
Nice work 👍

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+    for (let i = 0; i < this.length; i++) {
+      const length = this.length;
+
+      if (i === length - 1) {
+        result += this[i];
+        break;
+      }
+
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
```
varseniuk, line 19: Don't change the initial array. Now you change some element in `this` (in the initial array)

varseniuk (changes requested):

src/arrayMethodJoin.js
denyschemeriakov: Done)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const length = this.length;
+      let element = this[i];
+
+      if (i === length - 1) {
```
varseniuk, line 14:
```suggestion
      let element = this[i];

      if (i === this.length - 1) {
```
Actually, there's no need in this variable. You use it only once and you could just use `this.length` instead

varseniuk (approved):
Good!

PR #698, opened 2021-08-05
solution

VeronikaLushchik:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let newString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
```
denyschemeriakov, line 13: Don't change the initial array. Now you change some element in this (in the initial array) ©

That is what @varseniuk said me in the same case)

denyschemeriakov (changes requested):

Ihor-Karpyn (approved):
LGTM

PR #700, opened 2021-08-05
Solution

TarasHoliuk:


varseniuk (changes requested):
I see no student's review (

Andrii-Pinchuk (approved):
Good work 💪

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinResultString = '';
+
+    if (this.length < 0) {
+      return joinResultString;
+    }
```
```suggestion
    if (this.length < 0) {
      return joinResultString;
    }
```
It's redundant. You are returning `joinResultString` either way. If this has no length, the loop will not start and you eventually return the empty string

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+
+    let workSeparator = separator;
+
+    if (workSeparator === null) {
+      workSeparator = 'null';
+    }
```
```suggestion
```

I really have no idea, why would you need this. Just use `separator` from arguments

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      let element = this[i];
+
+      if (element === null || element === undefined) {
+        element = '';
+      }
+
+      if (separator !== undefined && this.length > 1 && i !== this.length - 1) {
```
varseniuk, line 27:
```suggestion
      if (i !== this.length - 1) {
```
the separator will never be undefined because you gave the default value for it.
The second check is aslo redundant

varseniuk (changes requested):
Let's remove some redundant things

src/arrayMethodJoin.js
TarasHoliuk: https://prnt.sc/1jt64yn
one of the tests doesn't work without this

varseniuk (approved):

PR #701, opened 2021-08-05
first commit

VVHimiak:


kristof22122 (approved):
Good job!

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
varseniuk, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```
You can give a default value to the argument instead of creating the `separator2` variable

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
...
-    // write code here
+    let separator2 = separator;
+
+    if (typeof separator === 'undefined') {
+      separator2 = ',';
+    }
```
```suggestion
```
Give default value to the argument

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
...
+
+    let testString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] === 'object' || typeof this[i] === 'undefined') {
```
varseniuk, line 17:
```suggestion
      if (this[i] === undefined || this[i] === null) {
```
You can just check if it's null or undefined

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
...
+        testString += `${separator2}`;
```
```suggestion
        testString += separator;
```
You can just add it without interpolation

varseniuk (changes requested):
Let's fix it a bit

varseniuk (approved):

PR #703, opened 2021-08-05
Method Join Solution

ihor-shapoval:


lera-yaskevych (approved):
Everything looks good to me 👌

danheim (approved):

PR #711, opened 2021-08-06
Solution

MaksymBeliever:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1) {
+        str += this[i];
+      } else if (this[i] === null || this[i] === undefined) {
+        str += ',';
```
illiarerun, line 14: I think you should use separator here
```suggestion
        str += separator;
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
```
illiarerun, line 8: Maybe rename variable to better one e. g. 'joinedString' or something else

illiarerun (approved):
Well done 🔥 Just needs a little bit of code polishing)

src/arrayMethodJoin.js
varseniuk: Illia is right. Really bad name for the variable

src/arrayMethodJoin.js
varseniuk: Yep

varseniuk (changes requested):
Let's fix naming and use Illia's suggestion

src/arrayMethodJoin.js
MaksymBeliever: Sure, I forgot to change that...

varseniuk (approved):

PR #715, opened 2021-08-06
solution arrayMethodJoin by Eugene Ghedenko

set1one:


roman-hado (approved):
Good job!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
```
varseniuk, line 12: Don't change the initial array. `join` method doesn't work like this
```suggestion
```

varseniuk (changes requested):
Don't change the array

VinnieJ-2k20 (approved):
Good job 👍

PR #718, opened 2021-08-06
js_array-method-join

RomanFedusevych:


DmytroMykoliv (approved):
Looks good )

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let newString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        this[i] = '';
```
```suggestion
```
Here you are changing the initial array. You shouldn't do this. Method `join` doesn't work like this

varseniuk (changes requested):
Don't change the initia array

VinnieJ-2k20 (approved):
Good job 👍

PR #720, opened 2021-08-06
Solution

Clonvhs:
v1.0

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,32 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedArray = '';
+
+    for (const item of this) {
+      if (item === null || item === undefined) {
+        joinedArray += '' + separator;
+      } else {
+        if (item !== this[this.length - 1]) {
+        }
+        joinedArray += item + '' + separator;
+      }
+    }
+
+    if (joinedArray[joinedArray.length - 1] === separator) {
+      joinedArray = joinedArray.slice(0, -1);
+    }
+
+    if (separator === null) {
+      joinedArray = joinedArray.slice(0, -4);
+    }
+
+    if (typeof separator === 'object' && separator !== null) {
+      joinedArray = joinedArray.slice(0, -15);
+    }
+
+    return joinedArray;
   };
 }
 
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^0.9.2",
     "eslint": "^5.16.0",
     "eslint-plugin-jest": "^22.4.1",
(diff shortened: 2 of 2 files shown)


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedArray = '';
+
+    for (const item of this) {
+      if (item === null || item === undefined) {
+        joinedArray += '' + separator;
+      } else {
+        if (item !== this[this.length - 1]) {
+        }
```
nastyyvolk, line 15: Something is supposed to be in this condition, right?

nastyyvolk (commented):
Good work so far!
I've left a few suggestions you may want to think about

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+        joinedArray += item + '' + separator;
+      }
+    }
+
+    if (joinedArray[joinedArray.length - 1] === separator) {
+      joinedArray = joinedArray.slice(0, -1);
+    }
```
nastyyvolk, line 22: As far as I understand your idea, this could be managed back in the 15th line

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedArray = '';
+
+    for (const item of this) {
+      if (item === null || item === undefined) {
+        joinedArray += '' + separator;
```
VinnieJ-2k20, line 12: joinedArray is already a string.
```suggestion
        joinedArray += separator;
```

src/arrayMethodJoin.js
VinnieJ-2k20, line 15: This code does nothing.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+      joinedArray = joinedArray.slice(0, -1);
+    }
+
+    if (separator === null) {
+      joinedArray = joinedArray.slice(0, -4);
+    }
+
+    if (typeof separator === 'object' && separator !== null) {
+      joinedArray = joinedArray.slice(0, -15);
+    }
```
VinnieJ-2k20, line 30: This check works, but it is too complicated.
Also, it won't work, if separator will be a string with `length > 1`

How about we change logic a little bit and just won't add separator after the last array element?

VinnieJ-2k20 (changes requested):
You are on the right way, but need to clean your code and simplify logic.

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,24 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedArr = '';
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null && this[i] !== []) {
```
Denchepornyuk, line 12: перевірка на пустий масив тут не потрібна `this[i] !== []` - ця умова завжди буде повертати true

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #721, opened 2021-08-06
Develop

Vova-Stoliar:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let arrayToString = '';
+    const cloneElements = this.concat();
+    const separatorToString = separator + '';
+
+    for (let i = 1; i < cloneElements.length; i += 2) {
+      cloneElements.splice(i, 0, separatorToString);
+    }
```
DmytroMykoliv: this task you can solved without methods 
        arrayToString += this[i] + `${separator}`;

DmytroMykoliv (approved):
It is working, but according to the condition you mast not use methods

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let arrayToString = '';
+    const cloneElements = [ ...this ];
```
VinnieJ-2k20, line 10: You don't mutate `cloneElements`. Why do you need this copy?

VinnieJ-2k20 (approved):
Good job 👍 
Just don't create unnecessary variables, if you don't have to mutate original array.

PR #724, opened 2021-08-06
solution

RomSmile:


LazurNiko (approved):
interesting solution!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const duplicateOfSeparator = separator + '';
+    let resultString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        this[i] = '';
```
VinnieJ-2k20, line 13: Don't mutate the original array. Default `join` method doesn't do it.

VinnieJ-2k20 (changes requested):
You need to avoid mutating original array for non-mutating methods.

VinnieJ-2k20 (approved):

PR #732, opened 2021-08-07
join done

yevhenpecherskyi:


Clonvhs (approved):
good one

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
...
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const input = [...this];
+    const inputToString = `${input}`;
+    let result = '';
+
+    if (input.length === 0) {
+      return '';
+    } else if (input.length === 1 || typeof separator === 'undefined') {
+      return `${input}`;
```
VinnieJ-2k20, line 15: You already have `inputToString` variable with same value.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
...
+    }
+
+    for (let i = 0; i < inputToString.length; i++) {
+      if (inputToString[i] === ',') {
```
VinnieJ-2k20, line 19: This won't work correctly if array contains strings wit commas.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const input = [...this];
```
VinnieJ-2k20, line 8: You don't mutate input array in your code. Why do you need this copy?

VinnieJ-2k20 (changes requested):
You are on the right way.
But some edits are still needed.
details in comments.

herkoss (approved):
Interesting solution!)

PR #761, opened 2021-09-13
solution

vmiskiv1:


KatyKuzmenko (approved):
Looks good!

Anna-Tkachenko (approved):
Good job!👍
You can also set the default value for separator in function arguments 'function(separator = ',')'
Did you check and fix lint errors?

PR #774, opened 2021-09-17
Solution

oksana-onopa:


KatyKuzmenko (approved):
Looks good!)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    if (this.length === 0) {
+      return '';
+    }
```
Anna-Tkachenko, line 12: If this.length === 0, cycle 'for' doesn't run and at the end of function will be returned empty string, so this if is redundant here

Anna-Tkachenko (approved):
Good job!👍

PR #778, opened 2021-09-17
solution

AVronskyi:


zhenia-verbitskaya (approved):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,11 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    if (separator !== undefined) {
+      return this.toString().replace(/,/g, separator);
+    } else {
+      return this.toString();
+    };
```
aliceblunt8, line 12: if you add default value for separator it could be one-liner ;)

```suggestion
      return this.toString().replace(/,/g, separator);
    };
```

aliceblunt8 (changes requested):
let's use algorithmic approach for solving this task (use loop)

aliceblunt8 (changes requested):
let's use algorithmic approach for solving this task (use loop)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
```
aliceblunt8, line 10: I believe you don't need it

aliceblunt8 (approved):

PR #782, opened 2021-09-17
--done

PresidentComanch:


Saharij:
Not bad!

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
+
+    if (this.length === 0) {
+      return str;
+    }
+
+    if (typeof separator === `undefined`) {
+      sep = ',';
+    } else {
+      sep = separator;
+    }
```
aliceblunt8, line 19: instead of this if else just add default value to separator argument

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
+
+    for (let el = 0; el < this.length - 1; el++) {
+      if (this[el] === null || this[el] === undefined) {
+        this[el] = '';
+      }
+      str = `${str}${this[el]}${sep}`;
```
aliceblunt8, line 25: or you can use string concatenation

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
+    }
+
+    str = str + this[this.length - 1];
```
aliceblunt8, line 28: let's do everything related to string concatenation inside the loop

aliceblunt8 (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+      if (this[el] === null || this[el] === undefined) {
+        this[el] = '';
+      }
+
+      if (el === this.length - 1) {
+        str = `${str}${this[el]}`;
+      } else {
+        str = `${str}${this[el]}${separator}`;
+      }
+    }
```
aliceblunt8, line 24:
```suggestion

      str = `${str}${this[el]}`;

      if (el !== this.length - 1) {
        str += separator;
      }
    }
```

aliceblunt8 (approved):

PR #784, opened 2021-09-18
solution

Bogdansgit:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,47 @@
...
+      case '' :
+        for (const el of this) {
+          joined += String(el);
+        }
+
+        return joined;
+
+      case undefined :
+        newSep = ',';
+    }
```
aliceblunt8, line 26: I think this switch case is unnecessary, just add default value to separator argument

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,47 @@
...
+        this[el] = '';
+      }
+
+      if (this[el] === false) {
+        this[el] = 'false';
+      }
+
+      if (this[el].toString === 'NaN') {
+        this[el] = 'NaN';
+      }
```
aliceblunt8, line 39: do you really need this? I think no

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,47 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let joined = '';
+    const joinedArr = [...this];
```
aliceblunt8, line 9: no need for separate variable, just use this

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,47 @@
...
+      if (this[el].toString === 'NaN') {
+        this[el] = 'NaN';
+      }
+
+      joinedArr[el] = (String(this[el]) + String(newSep));
+    }
+
+    for (const el of joinedArr) {
+      joined += el;
+    }
```
aliceblunt8, line 46: your solution seems too complicated, let's try to use only one loop in this task

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #803, opened 2021-09-20
solution

Mariia-Torkaienko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (const item of this) {
```
aliceblunt8, line 10: I think it's better to use for-loop here with index

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+      if (item === null || item === undefined) {
+        string += separator;
+        continue;
+      }
+
+      string += item;
+
+      if (this.lastIndexOf(item) !== this.length - 1) {
```
aliceblunt8, line 18: then you don't need lastIndexOf method

```suggestion
      if (i !== this.length - 1) {
```

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #814, opened 2021-10-07
solution

TRSKZK:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const incomingData = [...this];
```
MariiaObraztsova, line 8: Just curios: why did you create a copy instead of working with the original array?

VinnieJ-2k20: Yep, it's unnecessary here. You don't mutate `incomingData` array, so no need to make a copy of `this`.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+
+    if (incomingData.length > 0) {
+      let result = '';
+
+      for (let i = 0; i < incomingData.length; i++) {
```
VinnieJ-2k20, line 13: Don't write loops inside if. It's making your code too complicated.
Remember that loop with false condition simply will iterate 0 times.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+        element = '';
+      }
+
+      if (i === this.length - 1) {
+        result += element;
+        return result;
+      }
+      result += `${element}` + separator;
+    }
+    return result;
```
solaryasha, line 23:
```suggestion
			result += element;
			
      if (i !== this.length - 1) {
        result += separator;
      }
    }
    return result;
```

solaryasha (approved):

PR #818, opened 2021-10-09
join was created

misharosa:
=)

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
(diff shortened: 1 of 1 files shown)


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    // eslint-disable-next-line no-unused-vars
+    let word = '';
+
+    if (this.length === 0) {
+      return '';
+    }
```
DionStrzh, line 13: In this case you might return word as empty string

DionStrzh (changes requested):
It's great that you can read the tests, but you still must not use the methods here.
Even if tests allow to use `.concat()`

PR #868, opened 2021-11-09
Create custom join method

dmtrhrytsak:


dariia-romanova (approved):
Looks good!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const splitter = typeof separator === 'string' ? separator : separator + '';
```
illiarerun, line 8: You can use just String object constructor to avoid additional check, like String()

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+    let joined = '';
```
illiarerun, line 9:
```suggestion
    let joinedString = '';
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      const element = this[i] === null || this[i] === undefined ? '' : this[i];
+
+      joined += element;
+
+      if (i !== this.length - 1) {
+        joined += splitter;
+      }
```
illiarerun, line 18: Try to come up with solution where you don't need to check current index on each iteration

illiarerun (approved):
Great job 🔥

src/arrayMethodJoin.js
dmtrhrytsak: I just wanted to be precise like they all do in algorithms) But okay, I'll change that

PR #869, opened 2021-11-10
Join solution

Vitaliy-0:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
...
+    let workSep = separator;
+
+    if (workSep === null) {
+      workSep = 'null';
+    }
+
+    if (this.length) {
+      for (let i = 0; i < this.length; i++) {
+        if (this[i] === null || this[i] === undefined) {
+          this[i] = '';
```
WorkAccCom, line 18: I don't think, that changing the source array in this method is a good idea. Try to avoid it.

WorkAccCom (changes requested):
Good work, but need some refactoring.

src/arrayMethodJoin.js
Vitaliy-0: he don't modify the original array

aionova: You do modify the original array. Original join method doesn't do it.

aionova (changes requested):

aionova (approved):
GJ!

PR #874, opened 2021-11-10
Solution of method-join

Dima-Ziakhor:


junglq (approved):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,19 @@
...
-    // write code here
+    let joinedArray = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        joinedArray += this[i];
+      }
+
+      if (i < this.length - 1) {
+        joinedArray += separator !== undefined ? separator : ',';
```
aionova, line 16: this is not the best way to use ternary operator. You can give default value to the parameter separator instead

aionova (changes requested):

aionova (approved):

PR #875, opened 2021-11-10
solution

mariia-bila:


dmtrhrytsak (approved):
Switch doesn't go well with this type of task at all, come up with a solution where you don't use switch)

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,35 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let devider = separator;
+    let string = '';
+
+    switch (devider) {
+      case undefined:
+        devider = ',';
+        break;
```
illiarerun, line 14: For the undefined case, you can just use the default value which you can set in function declaration

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,35 @@
...
+      case null:
+        devider = 'null';
```
illiarerun, line 16: You can just use String() constructor, to convert such values to a string

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,35 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let devider = separator;
+    let string = '';
```
illiarerun, line 9: Be more specific with namings

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,35 @@
...
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      switch (this[i]) {
+        case null:
+          string += '';
+          break;
+        case undefined:
+          string += '';
+          break;
```
illiarerun, line 26: You can combine actions for these two cases

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,35 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let devider = separator;
```
illiarerun, line 8:
```suggestion
    let divider = separator;
```

illiarerun (changes requested):
Good start 🔥  Let's work on fixes now:

illiarerun (approved):
Great job 🔥

PR #880, opened 2021-11-11
Solution

olya-shyian:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const strSeparator = typeof separator === 'string'
+      ? separator
+      : separator + '';
```
dmtrhrytsak, line 10: It's 100% correct, but we can even make it shorter by always turning into a string, regardless of its type)
```suggestion
    const strSeparator = String(separator);
```

dmtrhrytsak: It won't break a program, only make it shine even more😎

dmtrhrytsak (approved):
Your join method is so good that I'm starting to think it could even join my life together, finally)
Nice going!

aionova (approved):
GJ!

PR #887, opened 2021-11-12
Solution

SerhiiKirik:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+
+    for (let index = 0; index < this.length; index++) {
+      const element = this[index] === null
+      || this[index] === undefined ? '' : this[index];
```
alex-kostomarov, line 12:
```suggestion
      const element = this[index] === null
        || this[index] === undefined ? '' : this[index];
```
i'm not sure, but it seems like there must be indented

and compound ternary operators are bit difficult to read

alex-kostomarov (approved):
Good solution!

src/arrayMethodJoin.js
SerhiiKirik: Webstorm told me, "Maaan, this is wrong, don't do this"
Do like this)0
If i use `ternary operators` here - then I don't need to do extra checks)

illiarerun (approved):
Good job 🔥  Check the comment below:

PR #888, opened 2021-11-12
Solution

alex-kostomarov:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let concatenatedStr = '';
+
+    for (let i = 0; i < this.length; i++) {
```
SerhiiKirik, line 10: What is `i`?
Ice? Idea? Image? Mb `Ilia`?)0

SerhiiKirik (approved):
Good job 🔥

src/arrayMethodJoin.js
illiarerun: Definitely Illia* 😂  But abstract indexes namings are not that scary, it's okay to use them

illiarerun (approved):
LGTM 🔥

PR #889, opened 2021-11-12
solution

Nadiyahr:


ImIkator (approved):
Good solution ! I'm not sure about checking separator on the beginning, it has to work without it. But everything done !

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
-    // write code here
+    let s = separator;
+
+    if (separator === null) {
+      s = 'null';
+    }
+
+    if (separator === undefined) {
+      s = ',';
+    }
```
aionova, line 16: you can do it by giving a default value to separator in the very beginning

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let s = separator;
+
+    if (separator === null) {
+      s = 'null';
+    }
```
aionova, line 12: can we write something else instead of this block?

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
+    if (separator === undefined) {
+      s = ',';
+    }
+
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      string += this[i] === null || this[i] === undefined
+        ? ''
+        : this[i];
```
aionova, line 23: this looks complicated. You can just write 1 "if" instead

aionova (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        string += '';
+      } else {
+        string += this[i];
+      }
+
```
aionova, line 18:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        string += this[i];
```

aionova (approved):

PR #892, opened 2021-11-12
Join

lera-permiakova:


tmrsl (approved):
Looks good!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        str += this[i];
+      }
+
+      if (this[i] !== this[this.length - 1]) {
+        str += separator;
+      }
```
ihor-jpeg, line 17: this is not the best solution here. in your case the method won't work correctly for all inputs.

ihor-jpeg: for example `[1,2,3,4,1].join2()` will yield `12,3,4,1` which is not correct. do you see why this is the case?

ihor-jpeg (changes requested):

src/arrayMethodJoin.js
lera-permiakova: got it! fixed

ihor-jpeg (approved):
Good stuff!

PR #899, opened 2021-11-14
Solution for join

Vitalii-Lytvyn01:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separationString = ',') {
+    let joinedString = '';
+    const separator = String(separationString);
+
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1) {
+        if (this[i] === null || this[i] === undefined) {
```
IrynaHerasymchuk, line 13: You don't need to use nested if conditions

IrynaHerasymchuk (approved):
Good! But think about simplifying your solution, you can actually solve this using only two if conditions😉

aionova (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separationString = ',') {
+    let joinedString = '';
+    const separator = String(separationString);
```
ihor-jpeg, line 9: when you concatenate a separator with a string, it will be converted to a string anyway

ihor-jpeg (approved):
Good work!

PR #901, opened 2021-11-15
Solution: join task

lanebx:


Veronika-Korona (approved):
Good job!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let newStr = '';
+
+    for (const i of this) {
+      if (i != null) {
```
ihor-jpeg, line 11: it's important to use strict equality. loose equality can behave in strange ways sometimes

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+        newStr += i;
+      }
+
+      if (this[this.length - 1] !== i) {
+        newStr += separator;
+      }
```
ihor-jpeg, line 17: what happens when the last element of the array is repeated somewhere else in the array. 
for example `[1, 2, 4, 3, 4].join2()`
can you predict why it won't work correctly?

ihor-jpeg (changes requested):
Good work. But let's fix a few things

src/arrayMethodJoin.js
lanebx: it was necessary to compare the indices. while i compare items.
If the element is the same as the last one, the separator will not work

corrected

src/arrayMethodJoin.js
lanebx: corrected

ihor-jpeg (approved):
well done!

PR #946, opened 2021-12-13
Create join

andrii-aleksandrov:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,34 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
     // write code here
+    let joined = '';
+    let separ = ',';
```
aliceblunt8, line 10: just add default argument value to the separator
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,34 @@
...
+      separ = separator + '';
+    }
+
+    if (!(this[0] === null || this[0] === undefined)) {
+      joined = this[0] + '';
+    }
+
+    if (this.length === 1) {
+      return joined;
+    }
```
aliceblunt8, line 22: remove these 3 if-blocks, you can do the same inside for-loop (just check for the last element)
if it's the last element in the array, you don't need to add separator after it (we can understand that the element is last if its index === this.length - 1)

aliceblunt8 (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joined = '';
+    const separ = separator + '';
```
aliceblunt8, line 9:
```suggestion
  [].__proto__.join2 = function(separator = ', ') {
    let joined = '';
```

aliceblunt8 (approved):

PR #958, opened 2021-12-16
join

Toros-Zz:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+    const newSeperator = '' + separator;
+
+    for (let i = 0; i < this.length; i++) {
+      let item = this[i];
+
+      if (item === undefined) {
```
marina-tilniak, line 14: lines 14 and 18 could be grouped, i think

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+        item = '';
+      }
+
+      if (item === null) {
+        item = '';
+      }
+
+      if (i === this.length - 1) {
+        str += item;
+        break;
```
marina-tilniak, line 24: maybe it makes sense just to use else?

Toros-Zz: I think it read better without 'else'

marina-tilniak (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        str += this[i];
+      }
+
+      if (i === this.length - 1) {
+        break;
```
Olexandr6, line 16:
```suggestion
      if (i !== this.length - 1) {
        str += `${separator}`;
```

Olexandr6 (approved):
Good job 👍  Check comments 👇

src/arrayMethodJoin.js
Toros-Zz: 👌😁

PR #970, opened 2021-12-17
Method-join

MVprogramm:


AlexAzalor (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (this.length === 0) {
+      return '';
+    };
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if ((this[i] === null) || (this[i] === undefined)) {
+        this[i] = '';
```
Elena-brv, line 16: do not assign smth to this[i]. You just shouldn't add it to the result string.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+      };
+
+      result += this[i] + '' + separator;
```
Elena-brv, line 19: why do you add an empty string?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+    };
+
+    if ((this[this.length - 1] === null)
+      || (this[this.length - 1] === undefined)) {
+      this[this.length - 1] = '';
+    };
+
+    return result + this[this.length - 1];
```
Elena-brv, line 27: would be good if the result is an ending value, without adding smth else

Elena-brv (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if ((this[i] !== null) && (this[i] !== undefined)) {
+        result += this[i];
+      };
+
+      if (i !== this.length - 1) {
+        result += String(separator);
```
Elena-brv, line 16: no need to convert it to String, the separator that is added to the string (result) will be converted to the string automatically

Elena-brv (approved):

PR #974, opened 2021-12-17
solution

samson-sim:


orest-mykhailyshyn (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        result += `${this[i]}`;
```
Elena-brv, line 12: when you add smth to string it will be converted to string also
```suggestion
        result += this[i];
```

Elena-brv (approved):

PR #983, opened 2021-12-18
solution_v1.0

Rozdorozhnii:


YevheniiKorzun (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+      return str;
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      const expression = (this[i] === null || this[i] === undefined)
+        ? ''
+        : this[i];
+
+      if (i !== this.length - 1) {
+        str += expression + String(separator);
```
Elena-brv, line 20: when you add string (expression) to anything on the separator (for example, a number) it will convert it to string automatically

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+    let str = '';
+
+    if (this.length === 0) {
+      return str;
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      const expression = (this[i] === null || this[i] === undefined)
+        ? ''
+        : this[i];
```
Elena-brv, line 17: a little complicated logic, maybe use this condition on if statement? 
if this condition is false, then we add this[i] to the str.

Elena-brv (changes requested):

src/arrayMethodJoin.js
Rozdorozhnii: In best practice of React developer often use ternary expression in their code, it's not looking too complicated. I'm practice in ternary expression here a bit.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+
+      if (i !== this.length - 1) {
+        str += expression;
+        str += separator;
+      } else {
+        str += expression;
+      }
```
Elena-brv, line 24: if you add expression in both cases, add it without any condition

Elena-brv (changes requested):

Elena-brv (approved):

PR #997, opened 2021-12-20
join 20.12

Rmu4h:


makssisua (approved):

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,33 @@
...
+
+    if (this.length === 0) {
+      return result;
+    }
+
+    if (element === undefined) {
+      element = ',';
+    } else {
+      element = separator + '';
+    }
```
Elena-brv, line 19: use default value for the parameter

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,33 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        this[i] = '';
```
Elena-brv, line 23: do not assign smth to this[i]. You just shouldn't add it to the result string.

Elena-brv (changes requested):

Elena-brv (approved):

PR #1004, opened 2021-12-25
join()

IvanShulhan:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,67 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let string = '';
```
aliceblunt8, line 8: add default value for separator in arguments like this `separator = ','`

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,67 @@
...
+
+    if (this.length === 0) {
+      return string;
+    }
+
+    if (this.length === 1) {
+      string += this[0];
+
+      return string;
+    }
```
aliceblunt8, line 18: remove these checks, they are reduntant

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,67 @@
...
+
+    for (let i = 0; i <= this.length - 1; i++) {
+      if ((this[i] === undefined || this[i] === null)
```
aliceblunt8, line 21: too many checks now
you only need to check for 2 main things
1) if this[i] !== null && this[i] !== undefined, then you can concatenate the next element
2) if this is not the last element of the string then you add the separator

that's all, try to refactor your code with these minimum checks in your mind

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #1008, opened 2021-12-28
Solution

KristinaTar:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const sep = separator === undefined ? ',' : separator;
```
aliceblunt8, line 8: just add default value to argument
like this
`function(separator = ',')`

aliceblunt8 (approved):

PR #1013, opened 2022-01-09
solution

ladrom:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,34 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let row = '';
+    let sep = String(separator);
+
+    if (separator === undefined) {
+      sep = ',';
+    };
```
aliceblunt8, line 13: add a default value to separator argument 
```
function(separator = ',') { ... }
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,34 @@
...
+          sep = '';
+        };
+        row += '' + sep;
+        continue;
+      };
+
+      if (typeof this[i] === 'object') {
+        row += '' + sep;
+        continue;
+      };
```
aliceblunt8, line 27: add this[i] to row only if this[i] is not null or undefined 
add separator separately in other if block if i !== this.length - 1

aliceblunt8 (changes requested):

src/arrayMethodJoin.js
ladrom: Если я все правильно понял, то сделал, хотя не уверен.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+      if (i === this.length - 1) {
+        sep = '';
+      };
+
+      if (typeof this[i] === 'undefined' || typeof this[i] === 'object') {
+        row += '' + sep;
+        continue;
+      };
+
+      row += this[i] + sep;
```
aliceblunt8, line 21:
```suggestion
    if (this[i] !== null && this[i] !== undefined) {
        row += this[i];
      }

     if (i !== this.length - 1) {
       row += `${separator}`;
     }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let row = '';
+    let sep = String(separator);
```
aliceblunt8, line 9:
```suggestion
```

src/arrayMethodJoin.js
aliceblunt8: the logic is that you add a current symbol and separator in different conditions one after another, do you understand now?

aliceblunt8 (changes requested):

src/arrayMethodJoin.js
ladrom: Yes, got it finally.

aliceblunt8 (approved):

PR #1023, opened 2022-01-12
solution

vince15ua:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,13 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const result = this.toString();
+
+    if (separator !== undefined) {
+      return result.replace(/,/g, separator);
+    } else {
+      return result;
+    }
```
aliceblunt8, line 14: try to come up with loop solution, without other methods

aliceblunt8 (changes requested):

danylo-onishchuk (approved):

PR #1030, opened 2022-01-20
solution

vsuslov29:


mtctxd (approved):
Гарна робота! По чеклісту все чисто. 🔥

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const correctedSeparator = separator;
```
Anna-Tkachenko, line 8: This variable is redundant, you don't change separator in this function so you can use argument variable

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if ([null, undefined].includes(this[i])) {
+        this[i] = '';
+      }
```
Anna-Tkachenko, line 17: Don't repeat yourself: you write two same if statements, one here and one in the next cycle

Anna-Tkachenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if ([null, undefined].includes(this[i])) {
+        this[i] = '';
+      }
```
Anna-Tkachenko, line 15: You can also try to do this with one cycle
Just add this if statement to the next one, and if it's true, add only separator to the result string and use continue to move to the next cycle iteration

Anna-Tkachenko (approved):

PR #1088, opened 2022-02-12
join method was implement

antonganetskiy:
done

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,30 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    let symbal = '';
+    let currentSep = separator;
+
+    if (separator === null) {
+      currentSep = 'null';
+    }
+
+    for (let i = 0; i < this.length; i++) {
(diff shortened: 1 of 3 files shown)


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    let symbal = '';
+    let currentSep = separator;
+
+    if (separator === null) {
+      currentSep = 'null';
+    }
```
SanyaBratashchuk, line 14: in my opinion that is unnecessary, you just can use an argument

SanyaBratashchuk (approved):

src/arrayMethodJoin.js
antonganetskiy: thanks

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    let symbal = '';
```
solaryasha, line 9:
```suggestion
    let symbol = '';
```

solaryasha (approved):

PR #1125, opened 2022-03-21
Solution

OlexandrIstomin:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
marina-tilniak, line 7: add default value for separator

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
-    // write code here
+    let newStr = ``;
+
+    if (this[0] === undefined || this[0] === null || this[0] === '') {
+      newStr += '';
+    } else {
+      newStr += `${this[0]}`;
+    }
```
marina-tilniak, line 14:
```suggestion
    if (this[0] !== undefined && this[0] !== null) {
      newStr += `${this[0]}`;
    } 
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null || this[i] === '') {
+        el = '';
+      } else {
+        el = this[i];
+      }
+
+      if (separator === undefined) {
+        newStr += ',' + el;
```
marina-tilniak, line 26: if you set default value then you don't need this if

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
+      } else if (separator === '') {
+        newStr += el;
```
marina-tilniak, line 28: this condition also redundant, next branch will work fine for empty string as well

marina-tilniak (changes requested):

marina-tilniak (approved):

PR #1127, opened 2022-03-23
Solution

serhii56465:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,31 @@
...
+
+    if (separator === null) {
+      str_ = str_.slice(0, -4);
+    } else if (typeof separator === 'object') {
+      str_ = str_.slice(0, -15);
+    } else if (separator !== '') {
+      str_ = str_.slice(0, -1);
+    }
+
+    return str_;
```
MasterpieceElbow, line 32: The solution is very complex and unreadable. I guess there is a more convenient solution.

MasterpieceElbow (changes requested):

PR #1156, opened 2022-04-04
Solution

serhii-naumenko:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,34 @@
...
+    let newString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        if (i !== this.length - 1) {
+          newString += ',';
+        }
+      } else {
+        i === this.length - 1 ? newString += this[i]
+          : newString += this[i] + separ;
```
YegorKochetkov, line 31:
```suggestion
        i === this.length - 1 
          ? newString += this[i]
          : newString += this[i] + separ;
```

YegorKochetkov (approved):
there are more elegant solutions, but it suits me)

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,35 @@
...
+          newString += ',';
+        }
+      } else {
+        i === this.length - 1
+          ? newString += this[i]
+          : newString += this[i] + separ;
+      }
+    }
+
+    return newString;
```
danylo-onishchuk, line 36: wow
so many if

its work, but let improve now


you need only 1 loop and 2 condition

danylo-onishchuk (changes requested):

danylo-onishchuk (approved):

PR #1158, opened 2022-04-04
Solution

Anastasiia-Svintsova:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
+    let result = '';
+    let sep = separator;
+
+    if (this.length === 0) {
+      return result;
+    }
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (sep === undefined) {
+        sep = ',';
```
VETALEK, line 17: you may just set the default value for `separator`

VETALEK (approved):
All fine

danylo-onishchuk (changes requested):
so many if else 

you need just 1 loop and 2 condition 

and use advise of VETALEK

danylo-onishchuk (approved):

PR #1159, opened 2022-04-05
join

Nikitiliok:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let trueSeparator = separator;
+
+    if (separator === undefined) {
+      trueSeparator = ',';
```
VETALEK, line 11: too complicated, you may just set default value for `separator` argument

VETALEK (approved):
Good, but a bit complicated

danylo-onishchuk (changes requested):
use advise of VETALEK

danylo-onishchuk (approved):

PR #1164, opened 2022-04-05
Apply join

vlasiuk-anatolii:


OleksandrNi (changes requested):
You can change one 'if' to 'switch')

See checklist
4. [CODE STYLE] - avoid using nested `if` statements, we are sure that you can do it without them :)

danylo-onishchuk (approved):

PR #1196, opened 2022-04-25
solution

efdalyalcin:


59MrRobot (approved):
Good job, I can't find room for improvement

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    const length = this.length;
+
+    let separUsed = separator;
+
+    if ('' + separUsed === 'undefined') {
+      separUsed = ',';
+    }
```
meforoma, line 15: better provide a default value for a function parameter instead this if block:
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
meforoma, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
+    let separUsed = separator;
+
+    if ('' + separUsed === 'undefined') {
+      separUsed = ',';
+    }
+
+    for (let i = 0; i < length; i++) {
+      if (typeof this[i] === 'undefined' || this[i] === null) {
+        ;
+        // if equal to just pass. I could not make !equal to work
```
meforoma, line 20: typeof is really not needed here
also you can use an implicit conditional - `undefined` and `null` are equivalent to `false` in conditionals
so `this[i] === 'undefined' || this[i] === null` === `!this[i]`
```suggestion
      if (!this[i]) {
        result += separator;
        // if equal to just pass. I could not make !equal to work
```

meforoma (changes requested):

meforoma (approved):
@efdalyalcin good call 👍 
> // !this[i] doesn't work since it removes 0 and NaN also

PR #1197, opened 2022-04-25
Solution

59MrRobot:


efdalyalcin (approved):
Good job!!

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    let newSeparator = separator;
+
+    if (separator === undefined) {
+      newSeparator = ',';
+    }
```
meforoma, line 13: this whole conditional block could be replaced by providing a default value right inside function parameters:
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
meforoma, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
...
-    // write code here
+    let result = '';
+    let newSeparator = separator;
+
+    if (separator === undefined) {
+      newSeparator = ',';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
```
meforoma, line 16: try implicit conditional:
- `null` and `undefined` are equivalent to false when used in conditional blocks
- so `this[i]` would replace your current conditional
   - and `!this[i]` would mean `this[i] === null || this[i] === undefined`

try it

meforoma (changes requested):

src/arrayMethodJoin.js
59MrRobot: I've tried using `this[i]` but it doesn't work

meforoma (approved):
yep @59MrRobot, since implicit conditional covers: `null`, `undefined`, `0` and `NaN`
it's handy to understand and keep in mind when working with conditionals

PR #1203, opened 2022-04-28
Solution

iLebo-stack:


andreisucman (approved):
Tbh I spent like 10 min and still couldn't quite understand how it works and why it works like that. But since it does the job and passed the tests I approve.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+    let joinedString = '';
+    let separatorCopy = separator;
+    const length = this.length;
+
+    for (let i = 0; i < length; i++) {
+      const element = this[i];
+
+      if (i === length - 1) {
+        separatorCopy = '';
+      }
```
VinnieJ-2k20, line 17: There is much easier way not to add separator after last element.

VinnieJ-2k20 (approved):
Good job 👍 
Check the comment.

PR #1208, opened 2022-04-28
Task Completed

MabuzaM:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedElements = '';
+    const thisLength = this.length;
```
rHashTag, line 9: Is this variable really necessary? You could just use `this.length` in the for loop.

rHashTag (changes requested):
Looks goods! I've added 1 comment to review.

src/arrayMethodJoin.js
MabuzaM: Yah I guess so, since I use it only once in the code. Thanx.

rHashTag (approved):
LGTM!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,18 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] === 'undefined' || this[i] === null) {
+        this[i] = '';
+      }
+
+      joinedElements += `${this[i]}${separator}`;
+    }
+
+    return joinedElements.substring(0, joinedElements.lastIndexOf(separator));
```
meforoma, line 18: smart 💡

meforoma (approved):

PR #1211, opened 2022-04-29
solution

SavMary:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,44 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+    let separ = separator;
+    const length = this.length - 1;
+
+    if (typeof separ === 'undefined') {
```
Anna-Tkachenko, line 12: Your separator can't be undefined, you set default value for it (coma) at function declaration
This if / else statement redundant, delete it

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,44 @@
...
+        case '':
+          break;
+        default:
+          if (i === length) {
+            str += this[i];
+          } else {
+            str += this[i] + separ;
+          }
+          break;
+      }
```
Anna-Tkachenko, line 41: This code looks too complicated and understandable
Try to optimize it
You should go throw your array and add elements + separator to result string
If it's the last element - separator is redundant (add one if for this case)

Anna-Tkachenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,34 @@
...
+            str += separ;
+          }
+          break;
+        default:
+          str += this[i];
+
+          if (i !== length) {
+            str += separ;
+          }
+          break;
```
Anna-Tkachenko, line 30: Will be better to use switch / case when you need to do some logic depending on some variables that has limited amount of values
In this case, if / else will be better option
Just check if element is null or undefined - add only separator
Else add element + separator (if it's not the last element)

Anna-Tkachenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+    let separ = separator;
+    const length = this.length - 1;
+
+    if (separator === null) {
+      separ = 'null';
+    }
```
Anna-Tkachenko, line 14: This check is redundant, try to delete it and pass the tests

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        if (i !== length) {
+          str += separ;
+        }
```
Anna-Tkachenko, line 20: Don't repeat yourself: you repeated this check 2 times, try to avoid it

Anna-Tkachenko (approved):

PR #1214, opened 2022-04-29
solution

vlad-demchuk:


SavMary (approved):
Exellent 👍🏻

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,37 @@
...
+      if (this[i] === undefined || this[i] === null) {
+        result += ',';
+        continue;
+      }
+
+      if (i === this.length - 1) {
+        result += String(this[i]);
+      } else {
+        result += String(this[i]) + separ;
+      }
```
Anna-Tkachenko, line 35: Don't repeat yourself (result += ...) : just add element to the result for all cases and add separator for all elements instead of the last

Anna-Tkachenko (changes requested):
Good job!
Just try to optimize your code

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separ = ',') {
```
Anna-Tkachenko, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

Anna-Tkachenko (approved):
Good job!

PR #1219, opened 2022-04-29
solution

alexandra-protyanova:


vlad-demchuk (approved):
Great job!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(sep = ',') {
```
Anna-Tkachenko, line 7: What is sep?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        this[i] = '';
+      }
+
+      if (i !== this.length - 1) {
+        result += this[i];
+        result += sep;
+      } else {
+        result += this[i];
```
Anna-Tkachenko, line 19: Don't repeat yourself (result += this[i]): you can move it outside of if / else statement

Anna-Tkachenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+    let separator2 = separator;
+
+    if (separator2 === null) {
+      separator2 = 'null';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        this[i] = '';
+      }
```
Anna-Tkachenko, line 18: You can also add needed separator here and use continue for moving to the next iteration

Anna-Tkachenko (approved):
Good job!

PR #1220, opened 2022-04-30
solution

yelnikov-andrii:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,29 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let res = '';
+    let separ = separator;
+
+    if (separator === undefined) {
+      separ = ',';
+    }
```
EnjSadman, line 13:
```suggestion
    [].__proto__.join2 = function(separator = ',') {
```
can usr this instead of if

EnjSadman (approved):

src/arrayMethodJoin.js
Anna-Tkachenko: Agree, set default value for separator in function declaration

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,29 @@
...
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        res += separ;
+      }
```
Anna-Tkachenko, line 18: Use key word continue here to go to the next iteration and avoid the next if / else statement

Anna-Tkachenko (changes requested):

src/arrayMethodJoin.js
yelnikov-andrii: it doesn't work if I use continue, because it needs to add a separator

Anna-Tkachenko: Add continue after the string with adding separator

Anna-Tkachenko (changes requested):
You didn't fix comments

Anna-Tkachenko (approved):

PR #1221, opened 2022-04-30
solution

lytvyshko:


iwanttosmile4u (approved):
Good job!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+    if (sep === null) {
+      sep = String(sep);
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      let item = this[i];
+
+      if (item === undefined || item === null) {
+        item = '';
+      }
```
Anna-Tkachenko, line 20: You can just add separator to the main string and use continue key word for moving to the next iteration

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+
+      if (i < this.length - 1) {
+        resultStr += item + sep;
+      } else {
+        resultStr += item;
+      }
```
Anna-Tkachenko, line 26: Don't repeat yourself (resultStr += item): add item for all cases but add separator for all elements except the last one

Anna-Tkachenko (approved):

PR #1222, opened 2022-04-30
make JOIN method

uliton:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,57 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const separatorValue = separator;
```
vlad-demchuk, line 8: try to set default value to arguments of the function

vlad-demchuk (changes requested):
try to optimize your code. there are many redudant checks and loops.  you can solve this task using only one for loop.

PR #1224, opened 2022-04-30
Solution

iwanttosmile4u:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultStr = '';
+    let separatorItem;
+
+    if (separator === null) {
+      separatorItem = 'null';
+    } else {
+      separatorItem = separator.toString();
```
valerii-derkach, line 14: it looks like you don`t need this if/else condition, because you do the same - you transform your separator to string in both cases;
-you can transform it either in declaration let separatorItem = ' ' + separator
- or you can  + it to you result string, but separatly in new line:
resultStr += this[i];
resultStr += separatorItem; - in this case do do not even need to assaign a variable for separator, just use income separator;
-or other method (like interpolation)
-or leave only first if (for null)

valerii-derkach (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        resultStr += separatorItem;
+      } else {
+        resultStr += this[i] + separatorItem;
+      }
+    }
+
+    return resultStr.slice(0, resultStr.length - separatorItem.length);
```
Anna-Tkachenko, line 19: Will be better to add some check in the cycle: if it's not the last element - add separator
If it's last element - don't add separator

Anna-Tkachenko (approved):

PR #1225, opened 2022-05-01
solution

valerii-derkach:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    let joiner = separator;
+
+    if (separator === undefined) {
+      joiner = ',';
```
iwanttosmile4u, line 12: You can set this value as default for separator in function arguments and don't use this checking.

Anna-Tkachenko (approved):
Good job!

PR #1229, opened 2022-05-01
First_commit_join

igoree88:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let newStr = ``;
+
+    for (let i = 0; i < this.length; i++) {
+      let oneWord = this[i];
```
iwanttosmile4u, line 11: This variable is redundant. Just use this[i] and don't overload your code with unnecessary variables.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+
+      if (oneWord === null
+        || oneWord === undefined
```
iwanttosmile4u, line 14: This is checking is also unnecessary.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+        || oneWord.length === 0) {
+        oneWord = '';
+      }
+
+      if (i === this.length - 1) {
+        newStr += `${oneWord}`;
```
iwanttosmile4u, line 20: Use this[i] instead of variable oneWord and you don't need such type of quotes as you use - ``. Also you can change this quotes into single quotes ' ' in newStr variable in the start.

iwanttosmile4u (approved):

Anna-Tkachenko (approved):
Good job!

PR #1230, opened 2022-05-02
join method is ready

Uktam19980416:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function (separator = ',') {
     // write code here
+    let joinMethod = '';
+
+    for (let idx = 0; idx < this.length; idx++) {
```
meforoma, line 11: naming is awkward
- `joinMethod` - i could think it's an actual method, but it's not - it's a `joinResult` | `resultString` etc., but clearly not a `Method`
- `idx` - not a common convention, use `i | j`

meforoma (changes requested):

VinnieJ-2k20 (approved):
Good job 👍

PR #1232, opened 2022-05-02
tasl is solved

enrikadej:


v-vytrykush (commented):
Good!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    if (this[0] !== undefined && this[0] !== null) {
+      string += this[0];
+    }
```
Anna-Tkachenko, line 12: This check is redundant here, just move it to the main cycle and check all elements on undefined or null

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job! 👍

PR #1235, opened 2022-05-03
Solution

v-vytrykush:


enrikadej (approved):
Perfect)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const thisLength = this.length;
```
Anna-Tkachenko, line 9: Why did you create separate variable for the length of array?

Anna-Tkachenko (approved):
Good job!

src/arrayMethodJoin.js
v-vytrykush: yeah, it is redundant here

PR #1239, opened 2022-05-03
solution

vitaliy-kovbasinskyi:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let NewString = '';
```
vlad-demchuk, line 8: don't name variables with uppercase letter

vlad-demchuk (approved):

src/arrayMethodJoin.js
Anna-Tkachenko: agree, fix this comment

Anna-Tkachenko (approved):

PR #1241, opened 2022-05-03
Solution

DNK-forexample:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+
+      if (this[i] === undefined && i === this.length - 1) {
+        continue;
+      }
+
+      if (this[i] !== undefined && i === this.length - 1) {
+        str += this[i];
+      } else {
+        str += this[i];
+        str += separator;
```
vlad-demchuk, line 25: don't repeat yourself (str +=...), you can add  all this[i] and do a check for the last separator

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
```
vlad-demchuk, line 8: it's better to use descriptive names of variables instead of 'str'

vlad-demchuk (approved):

Anna-Tkachenko (approved):
fix @vlad-demchuk comments

PR #1242, opened 2022-05-03
Solution

RomekIvantsiv:


vlad-demchuk (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (this.length === 0) {
+      return result;
+    }
```
Anna-Tkachenko, line 12: This check is redundant

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
+
+      result += `${this[i]}` + separator;
+    }
+
+    result += this[this.length - 1] ? this[this.length - 1] : '';
```
Anna-Tkachenko, line 22: You can also add the check in the cycle: if it's not the last element - add a separator

Anna-Tkachenko (approved):

PR #1244, opened 2022-05-04
solution

AndriiRut:
Solution

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,37 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
+
+    const separatorToString = '' + separator;
+    let result = '';
+
+    for (let i = 0; i < this.length - 1; i++) {
+      let char = this[i];
+
+      if (this[i] === undefined) {
+        char = '';
+      } else if (this[i] === null) {
+        char = '';
+      } else if (isNaN(this[i])) {
+        char = 'NaN';
+      }
+
+      if (this[i] === 'null') {
+        char = 'null';
+      }
+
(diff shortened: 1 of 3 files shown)


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,37 @@
...
+    let result = '';
+
+    for (let i = 0; i < this.length - 1; i++) {
+      let char = this[i];
+
+      if (this[i] === undefined) {
+        char = '';
+      } else if (this[i] === null) {
+        char = '';
+      } else if (isNaN(this[i])) {
```
dellachaise, line 22: why is NaN verification required?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,37 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
+
+    const separatorToString = '' + separator;
+    let result = '';
+
+    for (let i = 0; i < this.length - 1; i++) {
+      let char = this[i];
```
dellachaise, line 16: could you please write code from 16 line to 22 line(including 30 line) shorter?

maybe you can add char to string if it isn't null or undefined(using !==).

dellachaise (changes requested):
Good job!

Add please some changes according to comments.

src/arrayMethodJoin.js
AndriiRut: Thank you for advices!=)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
```
Anna-Tkachenko, line 10: This check is redundant

Anna-Tkachenko (approved):

PR #1246, opened 2022-05-04
Solution

rostyslav-pyliak:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
...
+    let joinedStr = '';
+    let separatorSymb = separator;
+
+    if (separator === undefined) {
+      separatorSymb = ',';
+    }
+
+    if (separator === null) {
+      separatorSymb = 'null';
+    }
```
vlad-demchuk, line 17: it'll be better to set default value to arguments of the function insteat of these checks

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
+
+      if (i === this.length - 1) {
+        joinedStr += this[i];
+      } else {
+        joinedStr += `${this[i]}${separatorSymb}`;
```
vlad-demchuk, line 27: try add all this[i] and do a check only for the last separatop, you'll avoid redundant checks in this way

vlad-demchuk (approved):

Anna-Tkachenko (approved):
⭐️

PR #1247, opened 2022-05-04
join

Oleksandr-Ivanchenko:


rostyslav-pyliak (approved):
Good

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let res = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let word = (this[i] === undefined || this[i] === null) ? '' : this[i];
+
+      if (i === this.length - 1) {
+        res += word;
+      } else {
+        res += word += String(separator);
```
Anna-Tkachenko, line 16: Don't repeat yourself: move res += word out of this if / else and add separator after all elements except the last

Anna-Tkachenko (approved):

PR #1250, opened 2022-05-04
join commit

volodya-leskiv:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let result = '';
+
+    if (this.length === 0) {
```
AndriiRut, line 11: You can stop function befor announcement of any variables.

AndriiRut (changes requested):

src/arrayMethodJoin.js
Anna-Tkachenko: This check is redundant

Anna-Tkachenko (approved):
⭐️

PR #1251, opened 2022-05-04
Method Join

Yaroslav-Kononenko:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    let char = String(separator);
```
uliton, line 9: separator allways is a string

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
...
+
+    if (separator === undefined) {
+      char = ',';
```
uliton, line 12: you can use it in default

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
...
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
```
uliton, line 17: you don't need this one if you fix my previous comment

src/arrayMethodJoin.js
Anna-Tkachenko: agree, set coma as default value for separator in function params

Anna-Tkachenko (approved):

PR #1265, opened 2022-05-19
develop

FANTAZER-NURE:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+        result += `${separator}`;
+      } else if (typeof this[i] === 'undefined' && i === this.length - 1) {
+        continue;
+      } else if (typeof this[i] === 'undefined') {
+        result += `${separator}`;
+      } else if (i === this.length - 1) {
+        result += `${this[i]}`;
+      } else {
+        result += `${this[i]}${separator}`;
+      }
```
aliceblunt8, line 21: I think 2 if-blocks should be enough
one to check whether the element is undefined or null and second to check the last element

aliceblunt8 (changes requested):

ArturErmolenko (approved):
Well done 👍

PR #1282, opened 2022-06-07
Solution

mira1zu:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+        result += separator;
+      }
+
+      element = this[i];
+
+      if ((element === undefined) || (element === null)) {
+        result += '';
+      } else {
+        result += element;
+      }
```
jstmpelowycz, line 22: in order to avoid adding `else`, you could place `element` declaration inside the loop

jstmpelowycz (approved):

PR #1284, opened 2022-06-07
Solution  Join

ghost:
Join mathod

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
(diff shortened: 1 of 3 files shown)


mflkvsk (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let exit = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        exit += this[i];
```
jstmpelowycz, line 12: don't you think `this[i]` looks like a new var?)

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #1289, opened 2022-06-08
solution

vinnibea:
Hate this one 😠

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,9 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (const element of this) {
+      if (this.indexOf(element) !== this.length - 1) {
+        result += typeof element === 'undefined' || element === null
+          ? '' + separator : element + '' + separator;
+      } else {
+        result += typeof element === 'undefined'
+          ? '' : element;
+      }
+    }
+
+    return result;
   };
-}
+};
 
 module.exports = applyCustomJoin;
--- a/package-lock.json
+++ b/package-lock.json
@@ -463,13 +463,248 @@
       }
     },
     "@mate-academy/scripts": {
-      "version": "0.3.5",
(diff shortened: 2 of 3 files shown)


jstmpelowycz (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,31 @@
...
+          result += element;
+          break;
+
+        case element === null || typeof element === 'undefined' :
+          result += '' + separator;
+          break;
+
+        default:
+          result += element + '' + separator;
+      }
```
illiarerun, line 27: that totally should not be a switch case. 
you repeat some conditions twice, you repeat some operations a lot of times, you should definitely redo it with if`s

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,31 @@
...
+        case this.indexOf(element) === this.length - 1:
+          result += element;
+          break;
+
+        case element === null || typeof element === 'undefined' :
+          result += '' + separator;
+          break;
+
+        default:
+          result += element + '' + separator;
```
illiarerun, line 26: you don't need this blank string here, like why do you even used it in the first place?

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,31 @@
...
+          result += '';
+          break;
+
+        case this.indexOf(element) === this.length - 1:
+          result += element;
+          break;
+
+        case element === null || typeof element === 'undefined' :
+          result += '' + separator;
+          break;
```
illiarerun, line 23: see here, you can add element first and then the separator to not repeat yourself

illiarerun (changes requested):
Nice try, now let's work on improvements

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,23 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        result += separator;
+      }
+
+      if (this[i] === null || typeof this[i] === 'undefined') {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
illiarerun, line 19: invert this condition so you wouldn't need to use `else`

illiarerun (changes requested):

PR #1290, opened 2022-06-08
Develop

Pavlo000:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let mySeparator = separator;
+    let join = '';
+
+    if (separator === undefined) {
+      mySeparator = ',';
+    }
```
jstmpelowycz, line 13: to avoid this, you could give `separator` a default value

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
...
+
+    for (let index = 0; index < this.length; index++) {
+      if (this.length - index > 1) {
+        join += `${
+          this[index] === null
+          || this[index] === undefined ? '' : this[index]}${mySeparator
+        }`;
+      } else {
+        join += this[index] === undefined ? '' : this[index];
+      }
```
jstmpelowycz, line 23: this structure is too complicated, I'm sure you could get it simpler

jstmpelowycz (changes requested):

Pavlo000:
I don’t know how to do it easier, maybe you can give some hint?

ср, 8 черв. 2022 р. о 17:10 Dmytro Pelovych ***@***.***> пише:

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+    let join = '';
+
+    for (const item of this) {
+      if (item !== undefined && item !== null) {
+        join += item;
+      }
+
+      if (this.indexOf(item) !== this.length - 1) {
+        join += separator;
+      }
```
jstmpelowycz, line 17: suggestion: I'd rather use for-with-indexes loop to avoid using `indexOf`, but still – up to you

jstmpelowycz (approved):

PR #1292, opened 2022-06-08
solution

Angelion541:


liliia-livandovskaya (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        joinedString += this[i];
+      }
```
jstmpelowycz, line 13: you should create a var to avoid duplicating `this[i]` every time – when there is a var, it's easier to maintain your code, you should get used to do it

jstmpelowycz (changes requested):

PR #1294, opened 2022-06-09
Solution

stanislav-topikha:


LesyaSvk (approved):
Wow.. looks interesting with ternary operator 🔥

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,17 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let el = this[i];
```
jstmpelowycz, line 11: what is `el`?)

jstmpelowycz (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,17 @@
...
+
+      el = (el === null || el === undefined) ? '' : `${el}`;
+      string += (i === this.length - 1) ? el : el + separator;
```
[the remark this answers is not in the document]
stanislav-topikha: blank line added

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,18 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let element = this[i];
+
+      element = (element === null || element === undefined) ? '' : `${element}`;
```
jstmpelowycz, line 13: using ternary, and then giving your var `''` – bad approach, you'd rather use _if_

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,18 @@
...
+
+      string += (i === this.length - 1) ? element : element + separator;
```
jstmpelowycz, line 15: ternary is handy, I love it, too
but let's look at this – as you can see, you duplicate `element`, is it right? again, simple _if_ could handle it better

jstmpelowycz: also, when you 100% sure ternary is a good idea, follow this code style approach
```
 something = condition 
  ? this 
  : that;
```

jstmpelowycz (changes requested):

src/arrayMethodJoin.js
stanislav-topikha: Changed a bit, but still using a ternary, because 1 of 2 need to be returned (direct purpose of ternary to return 1 of 2)

src/arrayMethodJoin.js
stanislav-topikha: Agree, repeating is bad, changed to `if`

illiarerun (approved):
GJ

PR #1311, opened 2022-06-09
Solution

MariiaNH:
Solution created

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let el = this[i];
+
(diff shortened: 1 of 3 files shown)


dimabelotskyi (approved):
GW!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let el = this[i];
```
jstmpelowycz, line 11: what is `el`?)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+
+      if (typeof el === 'undefined' || el === null) {
+        el = '';
+      }
+
+      result += el;
```
jstmpelowycz, line 17: revert this to avoid adding empty string

jstmpelowycz (changes requested):

PR #1322, opened 2022-06-10
solusion

DmytroSukhaniuk:


AndriiNychaiuk (approved):
Well done! 🔥

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let newChar = '';
```
jstmpelowycz, line 8: you don't return a char, you return a string, so next time name your vars carefully

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        newChar += this[i];
+      }
```
jstmpelowycz, line 13: `this[i]` looks like a new var

jstmpelowycz (approved):

PR #1325, opened 2022-06-10
Solution

aedricson:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,34 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    let currentSeparator = separator;
```
Olexandr6, line 9: You aren't gonna mutate `separator`, so you don't need to create a separate variable for it

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,34 @@
...
+
+    if (separator === undefined) {
+      currentSeparator = ',';
+    }
```
Olexandr6, line 13: Let's use the default function parameter instead

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,34 @@
...
+
+      if (i === this.length - 1) {
+        result += currentElement;
+        break;
+      }
+
+      result += `${currentElement}` + `${currentSeparator}`;
+    }
+
+    return result;
```
Olexandr6, line 35: Looks too complex, let's simplify this solution

Olexandr6 (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+      if (currentElement === null || currentElement === undefined) {
+        currentElement = '';
+      }
+
+      if (i === this.length - 1) {
+        result += currentElement;
+        break;
+      }
+
+      result += `${currentElement}` + `${separator}`;
```
Olexandr6, line 22:
```suggestion
      result += `${currentElement}${separator}`;
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
Olexandr6, line 8:
```suggestion
    let joinedString = '';
```

Olexandr6 (changes requested):

PR #1359, opened 2022-06-29
Solution

DarkMistyRoom:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,20 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
Bohdan-Maksymiuk, line 7: `function(separator = ',')`

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,20 @@
...
-    // write code here
+    let str = '';
+    const s = separator === undefined ? ',' : '' + separator;
```
Bohdan-Maksymiuk, line 9:
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,20 @@
...
+      let add = elem;
+
+      if (elem === null || elem === undefined) {
+        add = '';
+      }
+
+      str += add + s;
+    }
+
+    return str.substr(0, str.length - s.length);
```
Bohdan-Maksymiuk, line 21: just return `str` after small fix of logic in loop

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #1363, opened 2022-07-01
Solution

fledith:


silchyn (approved):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,38 @@
...
+      for (let i = 1; i < this.length; i++) {
+        str2 += `,${this[i]}`;
+      }
+
+      return str2;
+    }
+
+    if (typeof separator === 'number') {
+      return this;
+    }
```
VETALEK, line 24: do you really need this?

image

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,38 @@
...
+
+    if (separator === undefined) {
+      let str2 = this[0] + '';
+
+      for (let i = 1; i < this.length; i++) {
+        str2 += `,${this[i]}`;
+      }
+
+      return str2;
+    }
```
VETALEK, line 20: you have almost the same loop later in your code, you don't need to use it here
besides, here you don't prevent `null` and `undefined`

image

VETALEK (changes requested):
Nice start 🚀 

your code a bit too complicated

VETALEK (approved):
👍

PR #1376, opened 2022-07-06
Solution

Blankevich:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let sum = '';
```
valerii-derkach, line 8: for better code practice use semantic names please, it is not a `sum`, it is string or result, name it accordingly

valerii-derkach (changes requested):

src/arrayMethodJoin.js
Blankevich: corrected

alexandra-protyanova (approved):

valerii-derkach (approved):

PR #1377, opened 2022-07-06
Oleksandr-Pustovit

Oleksandr-Pustovit:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
...
+      switch (this[i]) {
+        case undefined:
+          break;
+
+        case null:
+          break;
+
+        case false:
+          result += 'false';
+          break;
```
VETALEK, line 20: `false` can be automatically casted to a string type

VETALEK (approved):
🔥

PR #1378, opened 2022-07-06
initial solution

DashaPoliazz:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,55 @@
...
+        const currentElement = this[i];
+        const isLastElement = i === this.length - 1;
+        const isElementValid = currentElement === undefined
+          || currentElement === null
+          || currentElement === '';
+
+        if (isLastElement) {
+          result += `${currentElement}`;
+        } else if (isElementValid) {
+          result += `INVALID_PROPERTY${separatorSymbol}`;
```
solaryasha, line 26: Why do you need to mark with `INVALID_PROPERTY` and then replace with empty space ?

solaryasha (changes requested):

solaryasha (approved):
Good job !

PR #1380, opened 2022-07-06
Solution

LeonidPustovit:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,39 @@
...
+
+          break;
+
+        case NaN:
+          joinedString += 'NaN';
+          break;
+
+        case false:
+          joinedString += 'false';
+          break;
```
VETALEK, line 28: these values can be automatically converted to string type

VETALEK (approved):
👍

PR #1383, opened 2022-07-06
join method solution

AlexanderKolomiiets:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
```
Githi54, line 11: In my opinion, better use typeof this[i] !== 'undefined'

danylo-onishchuk (approved):

PR #1387, opened 2022-07-07
solution

Kulchevych:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    if (arguments.length !== 0 && separator !== undefined) {
```
tiserett, line 8:
```suggestion
    if (arguments.length !== 0 && separator !== undefined) {
```
you don't need outer if statement

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
...
+      let resultStr = '';
+
+      for (let i = 0; i < this.length; i++) {
+        const item = (this[i] === undefined || this[i] === null)
+          ? '' : this[i];
```
tiserett, line 13:
```suggestion
        if (this[i] !== undefined && this[i] !== null) {
            resultStr += this[i];
        }
```
you don't need 'item' variable, you can check 'if this[i] !== undefined && this[i] !== null', if true, then resultStr += this[i]

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
...
+
+        if (i === this.length - 1) {
+          resultStr += item;
+        } else {
+          resultStr += item + `` + separator;
+        }
```
tiserett, line 19:
```suggestion
        if (i !== this.length - 1) {
          resultStr += separator;
        }
```
in my variant of code you don't need else statement, just add separator to the resultStr, you should also replace === with !==

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
...
+      }
+
+      return resultStr;
+    }
+
+    return this.toString();
```
tiserett, line 25:
```suggestion
      return resultStr;
    }
```
return only resultStr

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
tiserett, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```
Default value must be coma

tiserett (changes requested):
a few fixes required

tiserett (approved):
good job!

PR #1389, opened 2022-07-07
Solution

rodiontseva:


src/arrayMethodJoin.js
```diff
@@ -3,9 +3,33 @@
...
+//перебрать массив
+//если эл ок - добавить в новую строку
+//соследний эл доб без сапаратора
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let joinedItems = '';
+    let divider = separator;
+
+    if (separator === undefined) {
```
VETALEK, line 16: you may just set `,` as the default value for `separator`

VETALEK (approved):
🚀

src/arrayMethodJoin.js
rodiontseva: точно! спасибо большое, Виталик! пофиксила: https://github.com/mate-academy/js_array-method-join/pull/1389/commits/abbf757ac4daf795012c5445df44f4ba4bf912f4

PR #1392, opened 2022-07-07
Solution

Lentyaeev:
Initial commit

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,29 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    this.forEach((element, index) => {
+      let elementToAdd = element;
+      let separatorToAdd = separator;
+
+      if (separator === null) {
+        separatorToAdd = 'null';
+      }
+
+      if (elementToAdd === undefined || elementToAdd === null) {
+        elementToAdd = '';
+      }
+
+      if (index === this.length - 1) {
+        string += elementToAdd;
+      } else {
+        string += elementToAdd + separatorToAdd;
+      }
+    });
+
+    return string;
   };
 }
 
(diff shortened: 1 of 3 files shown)


illnino380 (approved):
I really like your solution. Using `forEach` makes the solution neater.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    this.forEach((element, index) => {
+      let elementToAdd = element;
+      let separatorToAdd = separator;
+
+      if (separator === null) {
+        separatorToAdd = 'null';
+      }
```
VETALEK, line 16: `separator` won't change during the function execution, so there is no need to check it on every step, you may do it only once before your `forEach`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+
+      if (elementToAdd === undefined || elementToAdd === null) {
+        elementToAdd = '';
+      }
+
+      if (index === this.length - 1) {
+        string += elementToAdd;
+      } else {
+        string += elementToAdd + separatorToAdd;
+      }
```
VETALEK, line 26: you add `elementToAdd` in each case, so there is no need to add it in `if` statement
```suggestion
      string += elementToAdd;
      
      if (index !== this.length - 1) {
        string += separatorToAdd;
      }
```

VETALEK (changes requested):
Good start 👍 
commented

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+    let separatorToAdd = separator;
+
+    if (separator === null) {
+      separatorToAdd = 'null';
+    }
+
+    this.forEach((element, index) => {
```
ivan-kadykalo, line 15: Try to make this task without array methods. Use for

ivan-kadykalo (changes requested):

ivan-kadykalo (approved):

PR #1394, opened 2022-07-08
Solution

illnino380:


Lentyaeev (approved):
A little bit complicated solution, but it works)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+        joinedItems += separator;
+        continue;
+      }
+
+      if (i === this.length - 1) {
+        joinedItems += this[i];
+        break;
+      }
+
+      joinedItems += this[i] + '' + separator;
```
VETALEK, line 25:
```suggestion
      joinedItems += this[i] + separator;
```

do you really need this `+ ''`?

VETALEK (changes requested):
GJ 🔥 
commented

src/arrayMethodJoin.js
illnino380: I've made it to complete the test:  if the separator is `Null`.

Ihor-Karpyn (approved):

PR #1403, opened 2022-07-10
solution

roman-romanenko:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let joinText = '';
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (typeof (separator) === 'undefined') {
+        joinText += this[i] + ',';
+      } else if (this[i] === null || typeof (this[i]) === 'undefined') {
+        joinText += ',';
```
yuliia-dronova, line 14: Maybe it is better to replace default value to the function's argument (separator = ',')

yuliia-dronova (approved):
Good!)

src/arrayMethodJoin.js
Ihor-Karpyn: agree, just add default separator

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
+      } else {
+        joinText += this[i] + String(separator);
+      }
+    }
+
+    if (typeof (this[this.length - 1]) !== 'undefined') {
```
Ihor-Karpyn, line 20:
```suggestion
    if (this.length > 0 && this[this.length - 1] != null) {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
+    let joinText = '';
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (typeof (separator) === 'undefined') {
+        joinText += this[i] + ',';
+      } else if (this[i] === null || typeof (this[i]) === 'undefined') {
+        joinText += ',';
+      } else {
+        joinText += this[i] + String(separator);
+      }
```
Ihor-Karpyn, line 17:
```suggestion
      if (this[i] != null) {
        joinText += `${this[i]}`;
      }

      joinText += `${separator}`;
```

Ihor-Karpyn (changes requested):

PR #1411, opened 2022-07-11
solution

vitmixG:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
+
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
```
Oleksandr-Pustovit, line 14: Don't you think first `if`  is extra here? 
Certainly, it is more of a question of clearness of code, but if you delete it, function should work just fine.
Your loop will not start ( `0 < (this.length = 0) => false`)  and `string = ''` will be returned.

Oleksandr-Pustovit (approved):
Perfectly done! :)
There is a small comment.

Ihor-Karpyn (approved):
LGTM

PR #1415, opened 2022-07-11
Solution_array-method-join

kata96451:


chernobrovin (approved):
Everything is fine. Good job!

PR #1419, opened 2022-07-11
array-method-join is solution

yuriiLevchenko2022:


kata96451 (approved):
only instead of "result", it is better to use more informative names, for example, "joined"

yuriiLevchenko2022:
> only instead of "result", it is better to use more informative names, for example, "joined"

yes, you are right!

PR #1430, opened 2022-07-21
Solution ver1.0

Oleksii-Korovnichenko:
Hi. 
Pllease, check my realization of method Array.join().
Thanks for review :-)


ArturErmolenko (approved):
well done 👍

PR #1433, opened 2022-07-22
Solution

mariabramnik:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,58 @@
...
+        stringResult += sep;
+        continue;
+      } else if (this[i] === false) {
+        stringResult += 'false';
+        stringResult += sep;
+        continue;
+      }
+
+      stringResult += this[i];
+
```
Bohdan-Maksymiuk, line 54: Good job, but it could be more simpler 
```
    let result = '';

        result += `${this[i]}`;
      }

      if (i < this.length - 1) {
        result += separator;
      }
    }

    return result;
  };
```

Bohdan-Maksymiuk (changes requested):

alexandra-protyanova (approved):

PR #1434, opened 2022-07-24
Solution

pelahiia:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (this.length < 1) {
+      result += this;
+
+      return result;
+    }
```
Bohdan-Maksymiuk, line 14: you don't need this

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        this[i] = '';
+      }
+
+      if (i === this.length - 1) {
+        result += this[i];
+      } else {
+        result += this[i] + String(separator);
+      }
```
Bohdan-Maksymiuk, line 25:
```suggestion
      if (this[i] !== undefined && this[i] !== null) {
        result += this[i];
      }

      if (i < this.length - 1) {
        result += separator;
      }
```
good, but better don't rewrite `this[i]`

Bohdan-Maksymiuk (changes requested):

marina-tilniak (approved):

PR #1438, opened 2022-07-26
Solution

maryna-mazur:


Bohdan-Maksymiuk (approved):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,48 @@
...
+            result += String(this[index]);
+            valueOfSeparator = ',';
+            result += valueOfSeparator;
+            break;
+
+          default:
+            result += String(this[index]);
+            result += valueOfSeparator;
+            break;
+        }
```
Bohdan-Maksymiuk, line 45: it all could be done more simpler
```
    let result = '';

        result += `${this[i]}`;
      }

      if (i < this.length - 1) {
        result += separator;
      }
    }

    return result;
  };
```

PR #1457, opened 2022-08-07
create metod join

KorniychukLesya:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultJoin = '';
+
+    for (let i = 0; i <= this.length - 1; i++) {
```
sergii-nosachenko, line 10:
```suggestion
    for (let i = 0; i < this.length; i++) {
```
It's not a mistake, but usually arrays are looped to values < length.

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,20 @@
...
+      if (this[i] !== null && this[i] !== undefined) {
+        resultJoin += `${this[i]}`;
```
sergii-nosachenko, line 12:
```suggestion
        resultJoin += this[i];
```
Unnecessary template literal. You need only variable to concatenate with string.

sergii-nosachenko (changes requested):
Good work. Some small fixes are needed.

KorniychukLesya:
Thanks for checking. I finished everything)

sergii-nosachenko (approved):
Nice 🔥

danylo-onishchuk (approved):

PR #1459, opened 2022-08-07
Solution

VolodymyrChuyko:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let str = '';
+    let currentSeparator;
+
+    if (separator === undefined) {
+      currentSeparator = ',';
```
Dmytro-Matsehora, line 12: you can get rid of the extra `if` construct here, if you add a default `separator` value

`function(separator = ',')`

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
...
+    } else {
+      currentSeparator = separator.toString();
+    }
+
+    for (const element of this) {
+      if (element === undefined || element === null) {
+        str += currentSeparator;
+      } else {
+        str += element + currentSeparator;
+      }
```
Dmytro-Matsehora, line 24: you can get rid of the extra `else` construct here, if you change the first condition a bit:

 ```
for (const element of this) {
      if (element !== undefined && element !== null) {
        str += element;
      }

      str += currentSeparator;
    }
```

Dmytro-Matsehora (changes requested):

Dmytro-Matsehora (approved):
good job! 👍

MaksimusCRS (approved):

PR #1470, opened 2022-08-10
Join- solution

NataliyaAn:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
Jamal-Amoian, line 7: You can add `separator = ','` as a default value

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
...
-    // write code here
+    let result = '';
+    const signSepar = (!separator && separator !== null && separator !== '')
+      ? ','
+      : separator;
+
+    if (this.length > 0) {
+      for (let i = 0; i < this.length - 1; i++) {
```
Jamal-Amoian, line 14: Check points 4 and 5 in the checklist

Jamal-Amoian (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,25 +4,18 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     let result = '';
```
Jamal-Amoian, line 8: What is it, result? please rewrite a name of your variable

Jamal-Amoian (changes requested):

Jamal-Amoian (approved):
Good job!

src/arrayMethodJoin.js
AlinaMatuschak: Agree with @Jamal-Amoian, result is bad name 💩

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined
+        && this[i] !== null) {
```
AlinaMatuschak, line 12: Also can create variable for it
```suggestion
      if (this[i] !== undefined && this[i] !== null) {
```

AlinaMatuschak (approved):
Nice, but should fix naming )

PR #1478, opened 2022-08-11
Solution

max-kravchenko:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,34 @@
...
+
+    let newSeparator = separator;
+
+    if (separator === undefined) {
+      newSeparator = ',';
+    }
+
+    if (separator === null) {
+      newSeparator = 'null';
+    }
```
sergii-nosachenko, line 19:
```suggestion
```

This part is unnecessary, cause you can set default value for separator directly in function parameters.

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,34 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
sergii-nosachenko, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```
Simpler and clearer way.

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,34 @@
...
+      if (this[i] === undefined || this[i] === null) {
+        this[i] = '';
+      }
+      str += this[i];
+
+      if (i === last) {
+        break;
+      }
+
+      str += newSeparator;
```
sergii-nosachenko, line 31: Here you can modify your condition to: `if i is not last - add separator`. Code will be shorter, cause you wouldn't need `break`.

sergii-nosachenko (changes requested):
Good job, but can be improved 😎

src/arrayMethodJoin.js
max-kravchenko: Great! Thank you!!

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,23 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+    const last = this.length - 1;
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
```
sergii-nosachenko, line 12:
```suggestion
      if (this[i] !== undefined && this[i] !== null) {
```

You can change the condition and move `str += this[i];` inside.

sergii-nosachenko (changes requested):

src/arrayMethodJoin.js
max-kravchenko: This helps me remove unnecessary 'kostyli')) Thank you!

sergii-nosachenko (approved):
Looks nice 🚀

danylo-onishchuk (approved):

PR #1481, opened 2022-08-11
Implement method join

Andrii-Stiebalov:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] !== 'undefined' && this[i] !== null) {
+        str += `${this[i]}`;
```
AlexanderKolomiiets, line 12: Why are you using interpolation here ?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+      }
+
+      if (i !== this.length - 1) {
+        str += `${separator}`;
```
AlexanderKolomiiets, line 16: here too

AlexanderKolomiiets (approved):
Great job! 
if you want to make you code cleaner, you can remove interpolation

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] !== 'undefined' && this[i] !== null) {
```
dellachaise, line 11: better simply check `this[i] !== undefined` in this case, because this way your code wold be cleaner.

Also you could read about difference between [variable === undefined vs. typeof variable === "undefined"](https://stackoverflow.com/questions/4725603/variable-undefined-vs-typeof-variable-undefined)
But it is now significant in this task.

dellachaise (approved):
🚀

PR #1482, opened 2022-08-11
Solution array method join

PastolNapas:


dellachaise:

OleksiiPustovit (approved):
Good job 👍

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let newSeparator = separator;
+
+    if (separator === undefined) {
+      newSeparator = ',';
+    }
```
dellachaise, line 12: No need to write this code, instead of it use default value for separator in line 7.

dellachaise (changes requested):
Great job 👍 

But better use default value for separator.

dellachaise (approved):
🚀

PR #1483, opened 2022-08-11
Solution

Anastasiia-Khmelovska:


AndreyJigijalkin (approved):
Great, but i'm not sure if you need extra test for string length in rows 10-12

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let newString = '';
+
+    if (this.length < 1) {
```
AlinaMatuschak, line 10: Unnecessary. Loop won’t run with an empty array

AlinaMatuschak (approved):
Nice work 🔥

PR #1485, opened 2022-08-13
Solution

plemiannik:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 1) {
+      return this[0].toString();
+    }
```
AlexanderKolomiiets, line 10: Why do you need this ?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+
+    let thisButStr = '';
```
AlexanderKolomiiets, line 12: give this variable more obvious name like `string` or `result`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        thisButStr += '';
+      } else {
+        thisButStr += this[i];
+      }
```
AlexanderKolomiiets, line 19: the same

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+
+      if (i < this.length - 1) {
```
AlexanderKolomiiets, line 21: in that case better use `!==` operator because you`re avoiding only one specific value.

AlexanderKolomiiets (changes requested):
great job, just add few fixes

src/arrayMethodJoin.js
plemiannik: > in that case better use `!==` operator because you`re avoiding only one specific value.

why? if i use i < this.length - 1 - i check every time is it and if i use i !== this.length - 1 i check every time is it

src/arrayMethodJoin.js
plemiannik: but everything can be string or result and from name thisButStr its obvious that it is this but string

src/arrayMethodJoin.js
plemiannik: cos no need in anything if this.length === 1

src/arrayMethodJoin.js
AlexanderKolomiiets: in this case you have only one variable which gonna return a result, so that's would be correct to name it `result `to make it obvious

plemiannik: nuok

src/arrayMethodJoin.js
AlexanderKolomiiets: `< operator ` would work too, but in my opinion, it's better to check only `this.length - 1` value
P.S. not `this.length` or `this.length + 3` and etc.

src/arrayMethodJoin.js
AlexanderKolomiiets: this if statement makes no sense, because if the length of this gonna be 1, so you would return that without separatator because of that rule.  
```
if (i < this.length - 1) {
       result += separator;
     }
```

AlexanderKolomiiets (changes requested):

src/arrayMethodJoin.js
plemiannik: but no need in cycle, no need in let result, no need in anything else, cos i can return what i search immiidiatly

src/arrayMethodJoin.js
plemiannik: i check only this.length - 1 value

AlexanderKolomiiets (approved):
nice solution, but still code could be shorter ))

plemiannik:
> nice solution, but still code could be shorter ))

is code must be shorter?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+    }
+
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
dellachaise, line 19: Try to rewrite with only `if` statement, without `else`. You need to add `this[i]` only in case  `this[i] !== null  and !== undefined`.

src/arrayMethodJoin.js
dellachaise, line 10: Better remove this line. Because for case with `this.length === 1` without this code it wouldn't be increasing algorithmic complexity. Because you cycle would have only one iteration, and creating variables - not so long process. But this way your code would be shorter.

dellachaise (changes requested):
Well done 👍 

Just add a few little fixes 👇

src/arrayMethodJoin.js
plemiannik: thnks, great idea, completely agree with u

dellachaise (approved):
Well done 👍

PR #1495, opened 2022-08-14
Method "Join"

Desertiks:


dellachaise:

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,24 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let arr = '';
+    const separatorCopy = separator + '';
+    // console.log(this);
+
+    for (let i = 0; i < this.length; i++) {
+      arr += (this[i] === undefined || this[i] === null ? '' : this[i])
```
Andrii-Stiebalov, line 12: if you want your code to be cleaner use if/else. For it is very difficult to understand what kind of logic you wanted to put into the ternary operator.

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,24 @@
...
+      + (i < this.length - 1 ? separatorCopy : '');
```
Andrii-Stiebalov: here also, it's better not to do calculations within calculations 🤯. Better use variables. You certainly won a few lines, but it turned into a mess 😄

Andrii-Stiebalov (approved):
I'm approving, but it will be better if you fix it 😉

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,16 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let arr = '';
+    const separatorString = separator + '';
```
dellachaise, line 9: Why do you need this variable, why don't you use separator?

dellachaise (changes requested):
Great star 👍 

Please add changes according to comments 👇

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+    let stringFromArray = '';
+
+    const endOfArray = this.length - 1;
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        stringFromArray += '';
+      } else {
+        stringFromArray += this[i];
+      }
```
dellachaise, line 17: No need in else statement. Because you add nothing.
```suggestion
      if (this[i] !== undefined || this[i] !== null) {
        stringFromArray += this[i];
      }
```

dellachaise (approved):
Well done 👍 

See comments 👇  for more optimal solution and fix it 😉

PR #1496, opened 2022-08-14
Solution

OlgaOchichenko:


YuriiPronin (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let arrayInString = '';
+
+    if (this === undefined || this === null || this.length === 0) {
+      return '';
+    }
```
dellachaise, line 12: This is unnecessary check, you could remove this code. Because this way only line 24 will execute even without check.

dellachaise (approved):
🚀

PR #1497, opened 2022-08-14
Develop

DimaKobzar7:


src/arrayMethodJoin.js
```diff
@@ -3,9 +3,24 @@
...
  */
+
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] === 'undefined' || this[i] === null) {
```
Reptiloid044, line 12: Good idea to use typeof!

Reptiloid044 (approved):
Nice code!
Good job!

DimaKobzar7:
Thanks for checking my work!

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,24 @@
 /**
  * Implement method join
  */
+
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
```
AlinaMatuschak, line 9: What is str? 💩

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,24 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] === 'undefined' || this[i] === null) {
+        str += '';
+      } else {
+        str += this[i];
+      }
```
AlinaMatuschak, line 16: if the element is undefined or null, you do nothing. So you can delete else

```suggestion
      if (typeof this[i] !== 'undefined' && this[i] !== null) {
        str += this[i];
      }
```

AlinaMatuschak (changes requested):
Almost done 😉

src/arrayMethodJoin.js
DimaKobzar7: String. I don't know what else there might be besides the string

AlinaMatuschak: Think more about it 🤓
You need to describe what this is. Just str not enough, I don't understand what it

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,22 @@
 /**
  * Implement method join
  */
+
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
AlinaMatuschak, line 9: result for what? This name doesn’t describe your variable

DimaKobzar7: function result

DimaKobzar7: and the result is a string

AlinaMatuschak: > and the result is a string
When I see result name, I don’t understand what is it

DimaKobzar7: I changed to stringOutcome. Is it good?

AlinaMatuschak: 😂
no

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,22 @@
 /**
  * Implement method join
  */
+
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringOutcome = '';
```
AlinaMatuschak, line 9:
```suggestion
    let joinedString = '';
```

AlinaMatuschak (approved):

PR #1498, opened 2022-08-14
Alternative method join was created

ArmPlay:
It works, but I don't like the cumbersomeness of my code. Tell me which is better?

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -7,25 +7,18 @@ function applyCustomJoin() {
   [].__proto__.join2 = function(separator = ',') {
     if (this.length === 0) {
       return '';
-    } else if (this.length === 1) {
-      return String(this[0]);
     }
 
-    let result;
-    const separatorToString = String(separator);
-
-    if (this[0] === undefined || this[0] === null) {
-      result = '';
-    } else {
-      result = this[0];
-    }
+    let result = (this[0] === undefined || this[0] === null)
+      ? ''
+      : String(this[0]);
 
     for (let i = 1; i < this.length; i++) {
       if (this[i] === undefined || this[i] === null) {
-        result += separatorToString;
+        result += separator;
         continue;
       }
-      result += separatorToString + String(this[i]);
+      result += separator + String(this[i]);
     }
 
     return result;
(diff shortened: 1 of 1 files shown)


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    } else if (this.length === 1) {
+      return String(this[0]);
+    }
```
stanislav-topikha, line 12: you can remove `else` part
```suggestion
    if (this.length === 0) {
      return '';
    }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
...
+
+    let result;
+    const separatorToString = String(separator);
+
+    if (this[0] === undefined || this[0] === null) {
+      result = '';
+    } else {
+      result = this[0];
+    }
```
stanislav-topikha, line 21: In both cases you are returning a value, ternary exists for this purposes
```suggestion

    const result = (this[0] === undefined || this[0] === null) ? '' : String(this[0]);
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    } else if (this.length === 1) {
+      return String(this[0]);
+    }
+
+    let result;
+    const separatorToString = String(separator);
```
stanislav-topikha, line 15: Do you really need to convert it to string?

stanislav-topikha (changes requested):
If you don't like 'cumbersomeness' try this ⬇️

src/arrayMethodJoin.js
ArmPlay: No need. Because even if it is a number, we still concatenate it with a string. Forgot to delete.

stanislav-topikha (approved):

PR #1502, opened 2022-08-14
solution

lizapolokhina:


dellachaise:

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,29 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let result = '';
```
nikolay-li, line 9: Better change variable name for something more informative.

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,29 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += separator;
+      } else {
+        result += this[i];
+        result += separator;
+      }
+    }
+
+    if (separator === '') {
```
nikolay-li, line 20: `  if (separator === '') {
      return (result);
    }

    if (separator === null) {
      return (result.slice(0, -4));
    }

    return result.slice(0, -separator.toString().length);`
    
    You can make this part short if check 'i' only at last iteration.
    
    But it's work, and let the mentor decide =)

nikolay-li (approved):

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,29 @@
...
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += separator;
+      } else {
+        result += this[i];
+        result += separator;
+      }
+    }
```
dellachaise, line 18: Better add `this[i]` to sting in case when  `(this[i] !== null || this[i] !== undefined)`.

And add one more `if` statement  for adding separator. Think what condition would you have.

This way code would be shorter and more understandable.

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,29 @@
...
+
+    if (separator === '') {
+      return (result);
+    }
+
+    if (separator === null) {
+      return (result.slice(0, -4));
+    }
```
dellachaise, line 26: And would be no need in this code.

dellachaise (changes requested):
Great start 👍 

Please add changes according to comments 👇

src/arrayMethodJoin.js
dellachaise: Better create more meaningful variables. For example `stringFromArray` instead of  `result`.

src/arrayMethodJoin.js
dellachaise, line 18:
```suggestion
    for (let i = 0; i < this.length; i++) {
      if (this[i] !== null || this[i] !== undefined) {
        result += this[i];
      }
      
      if (think what condition to use) {
        result += separator;
      }
    }
```

This way you could delete lines 20-26.

dellachaise (changes requested):

dellachaise (approved):
Well done 👍

PR #1504, opened 2022-08-14
solution

oksana-kyryienko:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
nikolay-li, line 8: Nice job!! =)
better change variable name to something more intuitive .

nikolay-li (approved):

src/arrayMethodJoin.js
AlinaMatuschak: Agree with @nikolay-li, result is bad name

AlinaMatuschak (approved):

PR #1506, opened 2022-08-15
Develop

OleksiiPustovit:


PastolNapas (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] !== 'undefined' && this[i] !== null) {
```
dellachaise, line 11: better simply check `this[i] !== undefined` in this case, because this way your code would be cleaner.

Also you could read about difference between [variable === undefined vs. typeof variable === "undefined"](https://stackoverflow.com/questions/4725603/variable-undefined-vs-typeof-variable-undefined)
But it is now significant in this task.

dellachaise (approved):
🚀

PR #1507, opened 2022-08-15
Solution

anna-kolmakova:


YuriiPronin (approved):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
```
AlinaMatuschak, line 8: What is result? Bad name 💩

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
+    let newSeparator;
+
+    if (separator === undefined) {
+      newSeparator = ',';
+    } else {
+      newSeparator = separator;
+    }
```
AlinaMatuschak, line 15: You can add default value
`function(separator = ',') {`

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
+
+    if (this.length === 0) {
+      return '';
+    }
```
AlinaMatuschak, line 19: Unnecessary. Loop doesn’t call with an empty array (you added `i < this.length` check)

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if ((this[i] === null)
+      || (this[i] === undefined
+        && (i !== this.length - 1))) {
```
AlinaMatuschak, line 24: It's hard to read

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
+        result += newSeparator;
+      } else if (this[i] !== undefined && i === this.length - 1) {
+        result += this[i];
+      } else if (i !== this.length - 1) {
+        result += this[i];
+        result += newSeparator;
+      }
```
AlinaMatuschak, line 31: you can use `if` without `if else`

```suggestion
      if (this[i] !== undefined && this[i] !== null ) {
        result += this[i];
      }
      
      if (i !== this.length - 1) {
        result += newSeparator;
      }
```

AlinaMatuschak (changes requested):
Almost perfect 👍

AlinaMatuschak (approved):
🚀🚀🚀

PR #1510, opened 2022-08-15
solution

roma-shafihullin:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] != null) {
```
David-Shahmatov, line 11: Я бы чисто для того чтобы спокойно спать , добавил бы ещё this[i] !== undefined =)

David-Shahmatov (approved):
Good job bro )

src/arrayMethodJoin.js
dellachaise: You could leave this code, but you should understand the difference between `!= null vs !== null`. In you case check to `undefined` already included. 
Reed more [here](https://stackoverflow.com/questions/38977829/what-is-the-difference-between-null-and-null)

dellachaise (approved):
🚀

PR #1515, opened 2022-08-15
Solution

Roman-Artiukh:


Andrii-Dunets (approved):
Well done)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let finallString = '';
+
+    for (const item of this) {
```
AlinaMatuschak, line 11: Maybe better use a simple for loop if you need index?

AlinaMatuschak (approved):
Good job 👏

src/arrayMethodJoin.js
Roman-Artiukh: Ok, i will change loop later, thanks))

PR #1516, opened 2022-08-15
solution

HusakOleh:


Reptiloid044 (approved):
Nice solution, good job!

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,26 @@
 /**
  * Implement method join
  */
+
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let res = '';
```
AlinaMatuschak, line 9: What is res? 💩

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,26 @@
...
+
+    if (this.length === 0) {
+      return '';
+    }
```
AlinaMatuschak, line 13: Unnecessary. I suppose that with an empty array loop won’t run

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,26 @@
...
+
+    for (const el of this) {
+      if (el !== null && el !== undefined) {
+        res += el;
+      }
+
+      if (this.indexOf(el, this.length - 1) !== this.length - 1) {
```
AlinaMatuschak, line 20: Maybe better use a simple for loop if you need index?

src/index.html
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
```
AlinaMatuschak, line 1: It shouldn’t be here

AlinaMatuschak (changes requested):
Good start, just small issues 👇

AlinaMatuschak (approved):
Awesome 🤩

PR #1526, opened 2022-08-19
Solution

vitaliidox:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    let separatorCopy = separator;
+
+    if (separator === undefined) {
+      separatorCopy = '';
+    } else if (separator === null) {
+      separatorCopy = 'null';
+    }
```
liliia-livandovskaya, line 15: you don't need this as you have already set the default value `separator = ',`'

liliia-livandovskaya (changes requested):
GJ! Pay attention to comments

src/arrayMethodJoin.js
vitaliidox: Ok. I`ve deleted 11-12 rows, but without next my code doesn`t pass tests. Or is that enough?)

valerii-derkach (approved):

PR #1535, opened 2022-08-26
Solution

AMahalias:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,37 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    let symbol = separator;
```
stanislav-topikha, line 9:
```suggestion
    let separatorCopy = separator;
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,37 @@
...
+    const copy = [...this];
```
stanislav-topikha, line 10:
```suggestion
    const arrayCopy = [...this];
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,37 @@
...
+
+    if (copy.length === 0) {
+      return '';
+    }
```
stanislav-topikha, line 14: its useless, cause if length = 0, result will not went to the loop

stanislav-topikha (changes requested):
comments ⬇️

stanislav-topikha (approved):
GJ 👍🏻

PR #1543, opened 2022-09-02
Solution

taras-nakonechnyi:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let finalString = '';
+    let newSeparator = separator;
+    let counter = 0;
+
+    if (separator === undefined) {
+      newSeparator = ',';
+    }
```
SanyaBratashchuk, line 14: set default value for parameter

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let finalString = '';
+    let newSeparator = separator;
+    let counter = 0;
```
SanyaBratashchuk, line 10: why do you need this counter? if you want to check index of char use a common `for` loop with increment

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
...
+      if (key === null || key === undefined) {
+        key = '';
+      }
+
+      if ((this.length - 1) === counter) {
+        finalString += `${key}`;
+      } else {
+        finalString += `${key}${newSeparator}`;
+        counter++;
+      }
```
SanyaBratashchuk, line 26:
```suggestion
      if ((this.length - 1) === counter) {
        finalString += `${key}`;
        break;
      }
        finalString += `${key}${newSeparator}`;
        counter++;
      
```

SanyaBratashchuk (changes requested):

stanislav-topikha (changes requested):

marina-tilniak (approved):

PR #1544, opened 2022-09-02
task solved

sergey-vorobei:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,27 @@
...
+      result += sep;
+    }
+
+    result += chek(this[this.length - 1]);
+
+    return result;
   };
 }
 
+function chek(element) {
```
qoa11a, line 22: Typo. check() is the correct variant

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,27 @@
...
+  if (element === undefined || element == null) {
```
qoa11a, line 23: Why not strict equality when comparing to null?

qoa11a (changes requested):

src/arrayMethodJoin.js
sergey-vorobei: corrected

src/arrayMethodJoin.js
sergey-vorobei: this is a typo.

alexandra-protyanova (approved):
GJ!

PR #1545, opened 2022-09-02
method Join

Dmitro-Chernolichenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,36 @@
...
+      return '';
+    }
+
+    if (copySource.length === 1) {
+      sep = '';
+    }
+
+    if (sep === null) {
+      sep = 'null';
+    }
```
Bohdan-Maksymiuk, line 22: it is not separator

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,36 @@
...
+    let result = '';
+    const copySource = [...this];
+
+    if (copySource.length === 0) {
+      return '';
+    }
+
+    if (copySource.length === 1) {
+      sep = '';
+    }
```
Bohdan-Maksymiuk, line 18: it can be 1 condition `<= 1`

Bohdan-Maksymiuk (changes requested):
you can use just 2 checks into loop
```
    let result = '';

        result += `${this[i]}`;
      }

      if (i < this.length - 1) {
        result += separator;
      }
    }

    return result;
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (this.length >= 0) {
```
SanyaBratashchuk, line 10: You don't need this check

SanyaBratashchuk (changes requested):

PR #1547, opened 2022-09-03
Solution

IhorChaikovskyi:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,23 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if ((i === (this.length) - 1) && this[i] === undefined) {
+        str += '';
+      } else if (i === (this.length) - 1) {
+        str += this[i];
+      } else if (this[i] === undefined || this[i] === null) {
+        str += separator;
+      } else {
+        str += this[i] + String(separator);
+      }
```
Bohdan-Maksymiuk, line 19: avoid such nested `if else` conditions

Bohdan-Maksymiuk (changes requested):

src/arrayMethodJoin.js
IhorChaikovskyi: Ok

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,28 @@
...
+        case (i === (this.length) - 1):
+          str += this[i];
+          break;
+        case (this[i] === undefined || this[i] === null):
+          str += separator;
+          break;
+        case true:
+          str += this[i] + String(separator);
+          break;
+      }
```
Bohdan-Maksymiuk, line 24: you can just use 2 if's instead of all this cases
```
        result += `${this[i]}`;
      }

      if (i < this.length - 1) {
        result += separator;
      }
```

Bohdan-Maksymiuk (changes requested):

src/arrayMethodJoin.js
IhorChaikovskyi: Ok, I understood that. Thank you

marina-tilniak (approved):

PR #1550, opened 2022-09-04
Solution

alexxxstep:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
Bohdan-Maksymiuk, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
+      if (!(this[ind] === null || typeof this[ind] === 'undefined')) {
+        element = String(this[ind]);
+      }
+
+      if (ind === this.length - 1) {
+        result = result.concat(element);
+        break;
+      }
+
+      result = result.concat(element, spr);
```
Bohdan-Maksymiuk, line 27:
```
        result += `${this[i]}`;
      }

      if (i < this.length - 1) {
        result += separator;
      }
```
you need only this here
it will be simplier

Bohdan-Maksymiuk (changes requested):

alexxxstep:
Дякую! виправив

Bohdan-Maksymiuk (approved):

PR #1554, opened 2022-09-06
Solution

MaxManis:


src/arrayMethodJoin.js
```diff
@@ -3,9 +3,40 @@
...
+// |-- but not an undefined, or NaN. XD
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let res = '';
+    let symbol = separator;
+
+    if (separator === null) {
+      symbol = 'null';
+    } else if (separator === '') {
```
Anna-Tkachenko, line 18: you can also set some value by default to the `separator` argument and avoid one of if / else statement

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,40 @@
...
  */
+
+// - There is kinda an easy solution for
+// |-- this task but with some weird tests
+// |-- like null could be a separator
+// |-- but not an undefined, or NaN. XD
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let res = '';
```
Anna-Tkachenko, line 13: what is res?

Anna-Tkachenko (changes requested):

src/arrayMethodJoin.js
MaxManis: done)
```
   funcName(separator = ',') {}
```

Denchepornyuk (changes requested):
Screenshot 2022-09-12 at 15 14 21

all good, but not perfect
here I tested your method join and original method. In your method something wrong with `NaN` as a separator and with numbers

Denchepornyuk (approved):

PR #1555, opened 2022-09-06
done

dmitriyn3679:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,22 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null && this[i] !== '') {
+        joinedStr += this[i];
+
+        if (i !== this.length - 1) {
+          joinedStr += separator;
+        }
+      } else if (i !== this.length - 1) {
+        joinedStr += separator;
```
Bohdan-Maksymiuk, line 18: avoid such nesting `else if` conditions

Bohdan-Maksymiuk (changes requested):

src/arrayMethodJoin.js
Bohdan-Maksymiuk: you may use 2 if's without nesting

PR #1560, opened 2022-09-09
alternativ method-join

RusNeday:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,38 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
Virass, line 7: set default value for separator

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,38 @@
...
+      switch (long > 0) {
+        case this[this.length - 1] === undefined && i === long - 1:
+          rezult += '';
+          break;
+
+        case (long - 1) === i || long <= 1:
+          rezult += this[i];
+          break;
+
+        case separator === undefined:
```
Virass, line 21: set default value for separator 
Like
`function(separator = ',')`

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,38 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let rezult = '';
```
Virass, line 8:
```suggestion
    let result = '';
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,38 @@
...
+    const long = this.length;
+
```
Virass, line 10: Add return for case when `this.length === 0` before loop

Virass (changes requested):
I think we have only 3 situation for join method
1. (this.length === 0)
2.Inside loop  (this[i] !== null && this[i] !== undefined)
3. Inside loop (i < this.length - 1)

Use this for refactoring your code

src/arrayMethodJoin.js
SanyaBratashchuk: Didn't fixed

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #1561, opened 2022-09-09
function arrayMethodJoin done

AnastasiaTitarenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+      }
+
+      if (this[i] !== undefined && this[i] !== null) {
+        if (i !== this.length - 1) {
+          str += `${this[i]}${separator}`;
+        } else {
+          str += `${this[i]}`;
+        }
+      }
+    }
```
Bohdan-Maksymiuk, line 24: avoid such nesting `else if` conditions

Bohdan-Maksymiuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -8,23 +8,35 @@ function applyCustomJoin() {
...
+  }
+}
+
+function lastElementCheck(el) {
+  if (el !== null && el !== undefined) {
+    return el;
+  } else {
+    return '';
+  }
+}
```
Virass, line 40:
```suggestion
function lastElementCheck(el) {
  if (el !== null && el !== undefined) {
    return el;
  }

  return '';
}
```

src/arrayMethodJoin.js
```diff
@@ -8,23 +8,35 @@ function applyCustomJoin() {
...
   };
 }
 
+function checkElement(el, separator) {
+  if (el === null || el === undefined) {
+    return separator;
+  } else {
+    return `${el}${separator}`;
+  }
+}
```
Virass, line 32:
```suggestion
function checkElement(el, separator) {
  if (el === null || el === undefined) {
    return separator;
  }

  return `${el}${separator}`;
}
```

Virass (changes requested):

PR #1566, opened 2022-09-12
Solution

SergIats:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,34 @@
...
+    }
+
+    const lastElement = this[this.length - 1];
+
+    if (lastElement === null || lastElement === undefined) {
+      string += '';
+    } else {
+      string += lastElement;
+    }
+
```
stanislav-topikha, line 34: Let's eliminate this, and make all check in the loop

stanislav-topikha (changes requested):

src/arrayMethodJoin.js
SergIats: Hi, Stanislav! You're right. Thanks! 
Done)

stanislav-topikha (approved):
GJ 👍🏻

PR #1567, opened 2022-09-12
Solution js_join_method

anton-martyniuk1:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let inFunctionSeparator = separator;
+
+    if (separator === null) {
+      inFunctionSeparator = 'null';
+    } else if (separator === '') {
```
qoa11a, line 12: How does it differ from row 8? It does the same thing, you don't need this condition

qoa11a (changes requested):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,27 @@
...
+    }
+
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        result += '';
+      } else {
+        result += this[i];
+      };
```
SanyaBratashchuk, line 21:
```suggestion
      if (this[i] !== undefined && this[i] !== null) {
       result += this[i];
      } 
```

SanyaBratashchuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let inFunctionSeparator = separator;
+    let result = '';
+
+    if (separator === undefined) {
+      inFunctionSeparator = ',';
+    }
```
liliia-livandovskaya, line 13:
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
liliia-livandovskaya, line 7:
```suggestion
  [].__proto__.join2 = function(separator=',') {
```

liliia-livandovskaya (changes requested):

liliia-livandovskaya (approved):

PR #1568, opened 2022-09-13
short solution without classic if

TheVovchik:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,17 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const initialLength = this.length;
+
+    let newStr = '';
+
+    for (let i = 0; i < initialLength; i++) {
+      newStr += (this[i] === undefined || this[i] === null) ? '' : this[i];
+      newStr += i !== initialLength - 1 ? separator : '';
```
grafit-off, line 14: Wow

grafit-off (approved):
GJ 🔥

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const arrayLength = this.length;
+
+    let convertedString = '';
+
+    for (let i = 0; i < arrayLength; i++) {
+      convertedString += (this[i] === undefined || this[i] === null)
+        ? ''
+        : this[i];
```
naziamoff, line 15: that's a nice one 👍 
you can improve it by creating separate variable for the condition

naziamoff (changes requested):
Great job! Just one small improvement

src/arrayMethodJoin.js
TheVovchik: ok, will do that

naziamoff (changes requested):
I meant separate boolean variable for the condition you've used in ternary operator
Let's say sth like that: 
`const isElementEmpty = element === undefined || element === null`
`convertedString += isElementEmpty ? '' : element`

naziamoff (approved):

PR #1574, opened 2022-09-14
Solution

glugovsky:


vmamchur (approved):
gj

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
```
OleksiiBlezniuk, line 8: What exact string is it? Create more meaningful variable name, it should be clear what exactly is stored in it

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      const item = (this[i] === null || this[i] === undefined)
+        ? ''
+        : this[i];
+
+      string += item;
```
OleksiiBlezniuk, line 15: It can be simplified, just don't add anything if element is null or undefined

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
👍

PR #1575, opened 2022-09-14
Solution

Andrew-Skoryk:
Solution

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let element = '';
+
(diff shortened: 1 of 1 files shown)


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      let element = '';
+
+      if (this[i] !== null && this[i] !== undefined) {
+        element = this[i];
+      }
+
+      string += element;
```
imaksweb, line 17:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        string += this[i];
      }
```

imaksweb (changes requested):

imaksweb (approved):
GJ 🔥

PR #1578, opened 2022-09-15
Solution

jojnajbude:


myroslav-diiak (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinStr = '';
```
OleksiiBlezniuk, line 8:
```suggestion
    let joinedStr = '';
```

OleksiiBlezniuk: join - action, joined - result of action, let's create more meaningful variables names

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+    const separatorStr = `${separator}`;
```
OleksiiBlezniuk, line 9: No need to create this separator string

OleksiiBlezniuk: If you'll add anything to string it will become string

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1 && this[i] !== null && this[i] !== undefined) {
+        joinStr += this[i];
+      } else if (this[i] !== null && this[i] !== undefined) {
+        joinStr += this[i] + separatorStr;
+      } else if (i === this.length - 1) {
+        break;
+      } else {
+        joinStr += separatorStr;
+      }
```
OleksiiBlezniuk, line 20: Too complicated logic, need to be refactored

OleksiiBlezniuk: Just add element to result string if it's not a `null` and `undefined`. Same with separator, just add it if element is not last

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):

PR #1579, opened 2022-09-15
js_array-method-join

xiimiik:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+    let resultStr = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const item = this[i];
+
+      if (item === undefined || item === null) {
+        resultStr += '';
+      } else {
+        resultStr += item;
+      }
```
imaksweb, line 17: You can simplify your `if else` statement:
```suggestion
      if (item !== undefined && item !== null) {
        resultStr += item;
      }
```

imaksweb (changes requested):

src/arrayMethodJoin.js
xiimiik: Of course, I didn't notice. Thank you!

imaksweb (approved):
Great job 👍

denys-cheporniuk (approved):

PR #1582, opened 2022-09-15
Custom implementation of Array.prototype.join()

Vladosik1337:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        str += this[i];
+      }
+
+      if (i !== this.length - 1) {
+        str += separator;
+      }
```
ArturNagaycev, line 17: does it work when we launch this function with argument null? I didn't pass the test. it's strange. ok, maybe I did something wrong

a-nahaitsev (approved):

src/arrayMethodJoin.js
Vladosik1337: image

ArturNagaycev: now agree. i understood when there was mistake in my code. thanks for comments ;)

naziamoff (approved):
gj

PR #1583, opened 2022-09-15
Solution

vladimir-dublya:


OleksandrGlulhovskiy (commented):
good job!

src/arrayMethodJoin.js
```diff
@@ -6,7 +6,18 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
     // write code here
+    let str = '';
+
+    if (separator !== undefined) {
+      str = this.toString().replace(/,/g, separator);
```
OleksiiBlezniuk, line 12: but what if some of array elements will have this commas?

src/arrayMethodJoin.js
```diff
@@ -6,7 +6,18 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
     // write code here
+    let str = '';
+
+    if (separator !== undefined) {
```
OleksiiBlezniuk, line 11: You can assign default value to function parameter

OleksiiBlezniuk (changes requested):
Let's implement this method without any methods (string methods too)

PR #1584, opened 2022-09-15
Solution

YuriMoreau:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    // write code her
+    let devider = separator;
+
+    if (devider === undefined) {
+      devider = ',';
+    }
```
SanyaBratashchuk, line 13: use default value for separator

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
+        result = result + this[i];
+        break;
+      }
+
+      if (this[i] === undefined || this[i] === null) {
+        result = result + devider;
+        continue;
+      }
+
+      result = result + this[i] + devider;
```
SanyaBratashchuk, line 31: You can simplify it
if element is not undefined or null concate it with result then if element is not a last concate separator. That's all what you need to do

SanyaBratashchuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        result = result + this[i];
+      }
+
+      if (i !== (this.length - 1)) {
```
SanyaBratashchuk, line 15:
```suggestion
      if (i !== this.length - 1) {
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        result = result + this[i];
```
```suggestion
        result += this[i];
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+      }
+
+      if (i !== (this.length - 1)) {
+        result = result + separator;
```
SanyaBratashchuk, line 16:
```suggestion
        result += separator;
```

SanyaBratashchuk (changes requested):

liliia-livandovskaya (approved):
Great job!

PR #1591, opened 2022-09-15
solution

myroslav-diiak:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,32 @@
...
   [].__proto__.join2 = function(separator) {
     // write code here
+
+    let separate = separator;
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    if (separator === undefined) {
```
jojnajbude, line 16: U don't need this condition, because u can set default value in parameters. Like this: function (separator = 0)

jojnajbude (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    let result = String(this[0]);
+
+    if (this[0] === null || this[0] === undefined) {
+      result = '';
+    }
```
OleksiiBlezniuk, line 18: Instead of this code, you can just create empty result string and iterate from first element instead of second in your loop. In case if array is empty - your loop wouldn't start and function will return empty string

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += separator + '';
+      } else {
+        result += separator + String(this[i]);
+      }
```
OleksiiBlezniuk, line 25: This logic can be simplified, just don't add anything if element is null or undefined. (adding empty string and not adding anything leads to same result)

OleksiiBlezniuk (changes requested):
Let's simplify your code

Also create more meaningful commit messages, it’s hard to understand what was done in this commit from `solution` message.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        result += String(this[i]);
```
OleksiiBlezniuk, line 13: Actually if you add anything to string it will be converted to string, but ok

PR #1592, opened 2022-09-15
1

vitaliy-beketov:


SanyaBratashchuk (changes requested):
too much redundant code 
All what you need to do is to set a default value to sepator then loop through array, if element is not an undefined or null concate it with result string, then if it is not a last element concate seperator

PR #1594, opened 2022-09-15
solution js_task-join

volodymyrklymkiv:


poznianski (commented):
gucci

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinResult = '';
+    const joinSeparator = String(separator);
```
Anna-Tkachenko, line 9: it's redundant here, you shouldn't create separate variable for separator, argument variable will be enough

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinResult = '';
```
Anna-Tkachenko, line 8: what is join Result? join it's an action, it can't be the name for the variable (it can be name for some function)

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,22 @@
...
+    const joinSeparator = String(separator);
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        this[i] = '';
+      }
+      joinResult += this[i] + joinSeparator;
```
Anna-Tkachenko, line 15: let's add check here, and will add separator only if it's not the last element

Anna-Tkachenko (changes requested):

PR #1596, opened 2022-09-15
Solution

AntonVaida:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (this.length < 1) {
+      return '';
+    }
```
illia-sosnytskyi, line 12: we'll have an `empty string` as the result with an empty array as a parametr anyway, so we don't need this `if` statement

illia-sosnytskyi (changes requested):

silchyn (changes requested):
image

silchyn (changes requested):
Try to stop modifying the array instead of creating a copy of the whole array

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
silchyn, line 15:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        result += this[i];
      }
```

silchyn (changes requested):

Anna-Tkachenko (approved):
🔥

PR #1597, opened 2022-09-15
solution

artem-musii:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
```

imaksweb (approved):
GJ!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joined = '';
```
Anna-Tkachenko, line 8: what is joined? let's create more meaningful variable name

Anna-Tkachenko (changes requested):

PR #1601, opened 2022-09-16
Solution

ghost:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,32 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    let sep = separator;
+
+    if (sep === undefined) {
+      sep = ',';
+    }
```
Yeliseiev, line 13: I think you forgot that you can assign a dafault value. Example: function(separator = ',')

Yeliseiev (approved):
GJ. Too much 'if' as for me, but still good work.

silchyn (changes requested):
Good point above

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const sep = separator;
```
silchyn, line 9: Do you need sep?

silchyn (changes requested):

silchyn (approved):
Cool!

PR #1603, opened 2022-09-16
solution

nikitaovch99:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,31 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = (',')) {
+    let result = '';
+
+    if (!this.length) {
+      return '';
+    }
```
imaksweb, line 12: I'm convinced you can easily get rid of this block it does nothing

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,31 @@
...
+
+    return result;
   };
+
+  function nullCheck(value) {
+    if (value === null || value === undefined) {
+      return '';
+    }
+
+    return value.toString();
```
imaksweb, line 30:
```suggestion
    if (value !== null && value !== undefined) {
      return value;
    }
```

imaksweb, line 30: If I'm not mistaken when you're trying to write something into a string it implicitly converts to string so there's no need to convert value to string. This process is going here:
```js
result += nullCheck(this[i]);
```

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,31 @@
...
+    if (!this.length) {
+      return '';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1) {
+        result += nullCheck(this[i]);
+      } else {
+        result += nullCheck(this[i]) + separator;
+      }
```
imaksweb, line 19: It looks more readable I suppose:

```suggestion
      result += nullCheck(this[i]);
      
      if (i !== this.length - 1) {
        result += separator;
      }
```

imaksweb (changes requested):

src/arrayMethodJoin.js
nikitaovch99: Agree, fixed.

src/arrayMethodJoin.js
nikitaovch99: Agree, fixed

src/arrayMethodJoin.js
nikitaovch99: Not so sure about this at least didn't work for me

imaksweb (approved):
👍

Anna-Tkachenko (approved):

PR #1611, opened 2022-09-16
Solution

Dikey945:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let joinedString = '';
+    let divider = separator;
+
+    if (divider === undefined) {
+      divider = ',';
+    }
```
imaksweb, line 13: This approach is deprecated, better to use default parameters:
```js
function(separator = ',')
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
...
+
+    if (this.length === 0) {
+      return joinedString;
+    }
```
imaksweb, line 17: This block of code does nothing, if you pass the empty string the loop is not executed and you just return joinedString, which you set as empty string on line 8

imaksweb (changes requested):

src/arrayMethodJoin.js
Dikey945: Agree, ty)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+    const divide = separator;
```
imaksweb, line 9: There's no need to use this constant at all 🙂

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      const element = this[i];
+
+      if (element !== undefined && element !== null) {
+        joinedString += `${element}`;
```
imaksweb, line 15: There's no need to use `template string`

```suggestion
        joinedString += element;
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+      }
+
+      if (i !== this.length - 1) {
+        joinedString += `${divide}`;
```
imaksweb, line 19: There's no need to use `template string` too
```suggestion
        joinedString += divide;
```

src/arrayMethodJoin.js
Anna-Tkachenko: agree, I don't see where you change this variable or some reason, why you can use one more variable for storing separator
so this variable is redundant, delete it

Anna-Tkachenko (approved):
Good job!

PR #1613, opened 2022-09-16
Solution

vallerii:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
```

imaksweb (approved):
👍

src/arrayMethodJoin.js
Anna-Tkachenko: agree, will be better to create other variable name, just it's not a good way to use data type in variable names

Anna-Tkachenko (approved):
Good job!

PR #1614, opened 2022-09-16
solution to method join

dsydorenko534:


iPanda177 (approved):
Nice work

silchyn (changes requested):
image

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let txt = '';
+
+    if (this.length === 0) {
+      return txt;
+    }
```
silchyn, line 12: Do you need it?

silchyn (changes requested):

silchyn (approved):
Cool!

PR #1615, opened 2022-09-16
Solution

varkovroma:


andrew-satyr (approved):
Well done)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const arrLength = this.length;
```
silchyn, line 8: Why we need that const?

silchyn (changes requested):

PR #1624, opened 2022-09-17
result

AndriyZhabynets:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
```
illia-sosnytskyi, line 10: we don`t need to check array length. it will work same way

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+
+    let firstEl = this[0];
+
+    if (firstEl === null || firstEl === undefined) {
+      firstEl = '';
+    }
+
+    let joined = `${firstEl}`;
```
illia-sosnytskyi, line 18: you can write it without interpolation

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
+
+      joined += `${separator}` + this[i];
```
illia-sosnytskyi, line 25: `you can write it without interpolation`
same here

illia-sosnytskyi (changes requested):

src/arrayMethodJoin.js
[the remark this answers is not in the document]
AndriyZhabynets: tests will not pass without interpolation

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let firstEl = this[0];
+
+    if (firstEl === null || firstEl === undefined) {
+      firstEl = '';
+    }
+
+    let joined = `${firstEl}`;
```
naziamoff, line 14: Why don't you check your first element inside the loop?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
```
naziamoff, line 18: you don't need to change this[i], you should modify only the result string

naziamoff (changes requested):

naziamoff (approved):
Nice!

PR #1629, opened 2022-09-18
Solution to Js_array-method-join

Fispil:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    let separatorCopy = separator;
+
+    if (typeof separatorCopy === 'undefined') {
+      separatorCopy = ',';
+    }
```
Bohdan-Maksymiuk, line 13: useless, you processed this case in 7 line

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+      if (typeof this[i] === 'undefined' || this[i] === null) {
+        this[i] = '';
+      }
+
+      if (i === this.length - 1) {
+        result += this[i];
+        break;
+      }
+
+      result += this[i] + `${separatorCopy}`;
```
Bohdan-Maksymiuk, line 30: * get rid of this breaks
* if `this[i] !== undefined && this[i] !== null` - add element to result
* if it is not the last element - add separator to result

Bohdan-Maksymiuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] === 'undefined' || this[i] === null) {
+        this[i] = '';
+      }
```
Bohdan-Maksymiuk, line 13: what do these lines do?
remove them
if (typeof this[i] === 'undefined' || this[i] === null) we will just add nothing to the result then

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #1632, opened 2022-09-18
Solution

roman-zherebetskyi:


AniaHna (commented):
Good job!
 😉

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separ = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const element = this[i];
```
silchyn, line 11: Why we need this const?

silchyn (changes requested):

src/arrayMethodJoin.js
roman-zherebetskyi: 'element' sounds better :grin:
okay I'll change, my bad...

silchyn (approved):
Cool!

PR #1634, opened 2022-09-18
Solution

rstarovoit:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
```
naziamoff, line 12: does method join change an original array?

naziamoff (changes requested):

src/arrayMethodJoin.js
rstarovoit: Nope) Fixed it!

naziamoff (approved):
Well done

PR #1635, opened 2022-09-18
Solution

tatyana22:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+    if (this.length === 0) {
+      return result;
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += `${this[i]}`;
+      }
```
avkovlit, line 20:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        result += this[i];
      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+
+      if (i < this.length - 1) {
+        result += `${separator}`;
```
avkovlit, line 23:
```suggestion
        result += separator;
```

avkovlit (approved):
GJ!
I left some comments, fix it please

src/arrayMethodJoin.js
silchyn, line 20:
```suggestion
      if (this[i]) {
        result += `${this[i]}`;
      }
```

silchyn (changes requested):
Good job, but

src/arrayMethodJoin.js
tatyana22: ok

src/arrayMethodJoin.js
tatyana22: ok

silchyn (approved):
Cool!

PR #1636, opened 2022-09-18
solution

Naveed-Shoukat:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,31 @@
...
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const arrayLength = this.length;
+
+    if (arrayLength === 0) {
+      return '';
+    }
+
+    const finalSeparator = separator === undefined
+      ? ',' : separator === null ? 'null' : separator;
```
Bibicutu, line 15: You could replace this long `if...else` chain with something else... Maybe with a `switch` statement.

Bibicutu (approved):
Good job!

src/arrayMethodJoin.js
ihor-jpeg, line 15: never nest ternaries. it's unreadable

use function parameters to assign a default value

ihor-jpeg (commented):
too complex

ihor-jpeg (changes requested):
.

src/arrayMethodJoin.js
Naveed-Shoukat: default value assigned to function parameter and nesting removed from ternary.

ihor-jpeg (approved):

PR #1639, opened 2022-09-18
v1: solution

igor-1987:


xiimiik (approved):
GJ 🔥

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let res = '';
```
Anna-Tkachenko, line 8: create more meaningful variable names

Anna-Tkachenko (changes requested):

src/arrayMethodJoin.js
igor-1987: It was just abbreviation :)

PR #1651, opened 2022-09-21
Solution for Join method

sapnachoudhary06:


kamal-tufekcic (approved):

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,27 @@
...
+      if (this[index] !== undefined && this[index] !== null) {
+        joined += this[index];
+      }
+
+      joined += separator;
+    }
+
+    if (this[this.length - 1] !== undefined && this[this.length - 1] !== null) {
+      joined += this[this.length - 1];
+    }
```
ihor-jpeg, line 25: what does this do?

ihor-jpeg (approved):

PR #1664, opened 2022-09-24
solution

Deniso4ka23:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+    const separator = args;
+    let result = '';
+
+    for (let i = 0; i < args.length; i++) {
+      if (separator[i] === undefined) {
+        separator[i] = ',';
+      } else if (separator[i] === null) {
+        separator[i] = 'null';
+      }
+    }
```
ArturErmolenko, line 17:
```suggestion
    let result = '';
```
you don't need this just give default value to separator

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(...args) {
```
ArturErmolenko, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+        result += this[i];
+      }
+
+      if (i < this.length - 1) {
+        result += separator;
+
+        if (!args.length) {
+          result += ',';
+        }
+      }
```
ArturErmolenko, line 30:
```suggestion
    if (this[i] !== null && this[i] !== undefined) {
         result += this[i];
       }

       if ((i !== this.length - 1)) {
         result += separator;
       }
      }
```

ArturErmolenko (changes requested):

ArturErmolenko (approved):
GJ 👍

PR #1666, opened 2022-09-25
Solution 1.0

aliliolek:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,60 @@
...
+    let sepSign = '';
+    let element = '';
+
+    if (separator) {
+      sepSign = separator;
+    } else if (separator === undefined) {
+      sepSign = ',';
+    } else if (separator === null) {
+      sepSign = 'null';
+    }
```
ArturErmolenko, line 19:
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,60 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
ArturErmolenko, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,60 @@
...
+      //       element = this[i];
+      //       break;
+      //   }
+      // }
+
+      if (i === (this.length - 1)) {
+        newString += element;
+      } else {
+        newString += element + sepSign;
+      }
```
ArturErmolenko, line 59:
```suggestion
 if (this[i] !== undefined && this[i] !== null) {
         newString += this[i];
       }

       if (i !== this.length - 1) {
         newString += separator;
       }
```

ArturErmolenko (changes requested):
you don't need to send on review task that isn't ready

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let newString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        newString += this[i];
+      } else {
+        newString += '';
+      }
```
TarasHoliuk, line 16:
```suggestion
    for (let i = 0; i < this.length; i++) {
      if (this[i] !== undefined && this[i] !== null) {
        newString += this[i];
      }
```

actually, `newString += '';` does nothing, so you can remove it

TarasHoliuk (changes requested):

polosanya (approved):
Good job!

PR #1680, opened 2022-10-07
solution

Hanna-Zaitseva:


SanyaBratashchuk (changes requested):
Let's simplify your code all what you need is to set a default value for separator, then iterate through array check if element is not null nor undefined and concate it to the result string, then check if it is not a last element and add seperator

marina-tilniak (approved):

PR #1681, opened 2022-10-09
Solution

Julia2063:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] !== 'undefined' && this[i] !== null) {
```
stanislav-topikha, line 11:
```suggestion
      if (char !== undefined && char !== null) {
```
let's make it work like this

stanislav-topikha (changes requested):

src/arrayMethodJoin.js
Julia2063: I have changed remarks, but i didnt use 'for of' loop, becouse i used index later, when i pointed to last element of array

polosanya (approved):
Good job!

PR #1682, opened 2022-10-09
Solution

Oleg-Lukanov:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let str = '';
+    let count = 0;
+
+    for (let ch of this) {
```
YevhS, line 12: Instead of using 'for of' loop, use 'for' loop with index access as it's required by the task description. Replace the loop with for (let i = 0; i < this.length; i++) {

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+      count++;
+
+      if (ch === undefined || ch === null) {
+        ch = '';
+      }
+
+      if (count !== this.length) {
```
YevhS, line 19: Replace count with i in the condition: if (i !== this.length - 1) {

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let str = '';
+    let count = 0;
+
+    for (let ch of this) {
+      count++;
```
YevhS, line 13: Since we will be using 'for' loop with index access, there's no need for 'count' variable. Remove this line.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
```
YevhS, line 7: It's better to set the prototype directly instead of using __proto__. Replace [].__proto__.join2 = function(separator = ',') { with Array.prototype.join2 = function(separator = ',') {

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let str = '';
+    let count = 0;
+
+    for (let ch of this) {
+      count++;
+
+      if (ch === undefined || ch === null) {
```
YevhS, line 15: Instead of checking if ch is undefined or null, you can check if this[i] is undefined or null.

YevhS (approved):

PR #1687, opened 2022-10-12
solution

mykola8426:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,20 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let string = this[0] === null || typeof this[0]
+    === 'undefined' ? '' : String(this[0]);
```
ArtemTopolskyi, line 9: what is going on here? lets rewrite it and make it readable for other developers

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,20 @@
...
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === null || typeof this[i] === 'undefined') {
+        string += separator + '';
+      } else {
+        string = `${string}${typeof separator === 'undefined'
+          ? ','
+          : separator}${this[i]}`;
```
ArtemTopolskyi, line 17: split logic into different steps, it's very hard to understand such code

ArtemTopolskyi (changes requested):

PR #1689, opened 2022-10-14
Solution

bokorzita:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,38 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const joinedArr = [];
+    let j = 0;
+    let newSeparator = separator;
+
+    if (newSeparator === undefined) {
+      newSeparator = ',';
```
amitpatiljc, line 13: let's use this as a default value for `seperator`

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,38 @@
...
+    }
+
+    for (let i = 0; i < (this.length - 1) * 2 - 1; i++) {
+      if (this[j] === null || this[j] === undefined) {
```
amitpatiljc, line 17: you have not defined `j` but you are using it here 
and you just need to implement normal join method, so require only normal for loop , you do not require to have loop till  `(this.length - 1) * 2 - 1`

amitpatiljc (changes requested):
for implementation of `Join` method you need to have just 2 conditions 
1 . as you already have for `undefined` or `null`  
2. for the last character as after that you do not required to add separator
and default you have to just add character and separator

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    const joinedArr = [];
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
+
+      if (i === this.length - 1) {
+        joinedArr[i] = this[i];
```
amitpatiljc, line 16: no need to have `else` statement as , if the `i` is last character then you have to add that character and break out of the loop 
so after adding last character use `break` keyword to end the process and break out of the loop

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+          joinedArr[i] = this[i] + separator;
+        }
+      }
+    }
+
+    let str = '';
+
+    for (const e of joinedArr) {
+      str += e;
+    }
```
amitpatiljc, line 30: instead of using loop to join array and convert it to string we can use `join` method

amitpatiljc (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const joinedArr = [];
```
jstmpelowycz, line 8: you'd rather give this array the following name
```suggestion
    const splitArray = [];
```
it consists of items of array

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
```
jstmpelowycz, line 13: you should not change the original array

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+      if (i === this.length - 1) {
+        joinedArr[i] = this[i];
+        break;
+      }
+
+      if (separator === null) {
+        joinedArr[i] = this[i] + 'null';
+      } else {
+        joinedArr[i] = this[i] + separator;
+      }
```
jstmpelowycz, line 24: you don't need it, `null` value is appended itself, gets converted into `"null"`

jstmpelowycz (changes requested):
almost done
generally, the approach is
```
...
  let output = '';

  for (...) {
    output += this[i] !== null && this[i] !== undefined 
      ? element 
      : '';

    if (index !== this.length - 1) {
      output += separator;
    }
  }
  ...
}
```

jstmpelowycz (approved):

PR #1691, opened 2022-10-15
commit

OleksChernikov:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,24 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
Bohdan-Maksymiuk, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,24 @@
...
-    // write code here
-  };
+    let string = "";
+    let division = separator;
```
Bohdan-Maksymiuk, line 9: no need

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,24 @@
...
+      }
+      if (division === null) {
+        division = 'null';
+      }
+      if (this[i] !== undefined && this[i] !== null) {
+        string += this[i];
+      }
+      if (this[i] !== this.length -1 && this.length > 1 && this[i] !== '') {
+        string += division;
+      }
```
Bohdan-Maksymiuk, line 22: do you need all these conditions here?
we may check only if current element is not the last one

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,24 @@
...
-  };
+    let string = "";
+    let division = separator;
+    for (let i = 0; i < this.length; i++) {
+      if (division === undefined) {
+        division = ',';
+      }
+      if (division === null) {
+        division = 'null';
+      }
```
Bohdan-Maksymiuk, line 16: join didn't add null as a separator if element is `null`

Bohdan-Maksymiuk (changes requested):

src/arrayMethodJoin.js
OleksChernikov: if delete conditions ( this.length > 1 && this[i] !== '') some tests fail.

polosanya (approved):
Good job!

PR #1694, opened 2022-10-17
Method join

maksym-kryvolap:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined && i === this.length - 1) {
+        result += '';
+      } else if (this[i] === null || this[i] === undefined) {
+        result += '';
```
tetiana-ivanchenko, line 14: this action do nothing, why do you need it?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined && i === this.length - 1) {
+        result += '';
```
tetiana-ivanchenko, line 12: this condition is redundant, why do you need it?

src/arrayMethodJoin.js
tetiana-ivanchenko: could you do a little brainstorming to avoid redundant conditions? Actually you can solve this task with only two "if" conditions. The first time you need to make sure your current element is not null and undefined, and if not, add it to the result. And then you just have to check if your element is not the last one and add a delimiter if it isn't.

tetiana-ivanchenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
```
TarasHoliuk, line 12: Good improvement!
But no need to add an empty string (`''`). just check the case where you want to add smth (use `!==`)

TarasHoliuk (changes requested):

TarasHoliuk (changes requested):

marina-tilniak (approved):

PR #1695, opened 2022-10-19
commit

devWizHub:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
     // write code here
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        str += this[i];
+      }
+
+      if (i < this.length - 1) {
+        str += separator;
```
amitpatiljc, line 17: but what if it is `i === this.length - 1` ?

amitpatiljc (changes requested):

src/arrayMethodJoin.js
devWizHub: code changed.

amitpatiljc (approved):
Good Job

PR #1704, opened 2022-10-23
Array method solution

karyna-salnykova:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+    const arrLength = this.length;
+    let joinStr = '';
+
+    for (let i = 0; i < arrLength; i++) {
+      if (typeof this[i] === 'undefined' || this[i] === null) {
+        this[i] = '';
+      }
+
+      if (i !== arrLength - 1) {
+        joinStr += this[i] + String(separator);
```
masha-girya, line 17: i see String(separator) second time in review, it is interesting for me why you needed it? Is it nessecary?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const arrLength = this.length;
+    let joinStr = '';
+
+    for (let i = 0; i < arrLength; i++) {
+      if (typeof this[i] === 'undefined' || this[i] === null) {
+        this[i] = '';
+      }
```
masha-girya, line 14: here you can write the loop whithout if-else, you just can start from if this[i] is NOT undefined or null :)

masha-girya (approved):
good job))

src/arrayMethodJoin.js
illiarerun: Agree, why don't you just add this[i] to string if it's not null or undefined, you're repeating yourself in next if block - try to invert your if statement to do that. Also, you don't need typeof here.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+
+      if (i !== arrLength - 1) {
+        joinStr += this[i] + String(separator);
+      } else {
+        joinStr += this[i];
+      }
```
illiarerun, line 20: and then you'd only add separator here like that
```suggestion
      if (i !== arrLength - 1) {
        string += String(separator);
      } 
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const arrLength = this.length;
+    let joinStr = '';
```
illiarerun, line 9: don't be shy with your variable names
```suggestion
    let joinedString = '';
```

illiarerun (changes requested):

denys-danyliuk (approved):

PR #1705, opened 2022-10-23
join Masha Girya

masha-girya:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let string = '';
+
```
Stepico, line 10: I guess this name is bad, just string doesn't get us any understanding of what this string is created for

Stepico (approved):
Good job, just one little piece of advice from me

src/arrayMethodJoin.js
denys-danyliuk: agree

denys-danyliuk (approved):
GJ!

PR #1706, opened 2022-10-24
Solution

andriy-shymkiv:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+    const length = this.length;
+
+    for (let i = 0; i < this.length; i++) {
```
iryna-lelyk, line 11: Why don't you use here your variable 'length'?

iryna-lelyk (approved):
Looks good to me.

src/arrayMethodJoin.js
Andriy-Shymkiv: good question ))

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+    const length = this.length;
```
illiarerun, line 9: don't create redundant variables
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
```
illiarerun, line 8: don't use names like that, try to come up with more meaningful name

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+    const length = this.length;
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] === 'undefined'
+      || this[i] === null) {
+        this[i] = '';
+      }
```
illiarerun, line 15: Why don't you just add `this[i]` to string if it's not `null` or undefined, you're repeating yourself in next `if` block - try to invert your if statement to do that. Also, you don't need `typeof` here.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+
+      if (i !== length - 1) {
+        string += this[i] + String(separator);
+      } else {
+        string += this[i];
+      }
```
illiarerun, line 21: and then you'd only add separator here like that
```suggestion
      if (i !== length - 1) {
        string += String(separator);
      } 
```

illiarerun (changes requested):

illiarerun (approved):

PR #1707, opened 2022-10-24
Solution was Created

Stepico:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+    for (const item of this) {
+      if (item === null || item === undefined) {
+        joinedString += '' + separatorToString;
+      } else {
+        joinedString += item + separatorToString;
+      }
+    }
+
+    // eslint-disable-next-line max-len
+    return joinedString.slice(0, joinedString.length - separatorToString.length);
```
masha-girya, line 20: a little bit complicated as I like)))) but mentors dont:(

masha-girya (approved):

src/arrayMethodJoin.js
Stepico: The separator's length can be more than 1, so in this case I decided that I need exactly separator.length instead of just 1

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+    const separatorToString = String(separator);
+
+    for (const item of this) {
+      if (item === null || item === undefined) {
+        joinedString += '' + separatorToString;
+      } else {
+        joinedString += item + separatorToString;
+      }
```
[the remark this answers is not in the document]
illiarerun: Agree, why don't you just add item to string if **ITS NOT** null or undefined? And only after that add your separator. Also, I don't like this `slice` thing in the end, let's just not add separator after last element? Try to use indexed for

illiarerun (changes requested):

illiarerun (approved):

PR #1709, opened 2022-10-24
Method join was added

iryna-lelyk:


Illia-Skladnik (approved):
Nothing to add here. Agree with decision.

denys-danyliuk (approved):
🔥

PR #1711, opened 2022-10-24
solution

taraselnik:


Svitlana-Yudina (approved):
Great Job! :fire:

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
illiarerun, line 8: let's not use names like `result` or smth, try to come up with more meaningful names

illiarerun (approved):

PR #1719, opened 2022-10-25
implemented method join

Artem66:


Largoxoxo (approved):
Gj!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let separatedString = '';
+    const lastIndex = this.length - 1;
+
+    for (let i = 0; i < lastIndex; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        separatedString += `${separator}`;
+      } else {
+        separatedString += `${this[i]}${separator}`;
+      }
```
illiarerun, line 16: split this if in two:
- one to check if you need to add the element, do it ONLY if it's NOT null or undefined
- second one to check if you need to add the separator, do it ONLY if you're NOT working with the LAST element

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+    }
+
+    if (this[lastIndex] !== null && this[lastIndex] !== undefined) {
+      separatedString += `${this[lastIndex]}`;
+    }
```
illiarerun, line 21: then you wouldn't need this check at all

illiarerun (changes requested):

mykola-koval (approved):

PR #1723, opened 2022-10-25
complete

spiritkyiv:


Anastasiia-Smirnova (approved):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let res = '';
+    const stringSep = typeof (separator) === typeof (undefined)
+      ? ','
+      : String(separator);
```
illiarerun, line 11: instead of doing this, use default params for your function
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
illiarerun, line 7: like that
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
-    // write code here
+    let res = '';
```
illiarerun, line 8: res is a bad variable name, try to come up with smth better, like `joinedString`

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
+
+    for (let x = 0; x < this.length; x++) {
+      if (this[x] === undefined || this[x] === null) {
+        this[x] = '';
+      }
+
+      res += this[x]
+      + (x === this.length - 1
+        ? ''
+        : stringSep);
```
illiarerun, line 21: instead of adding this weird ternary operator, better invert your first if statement and add element is it's NOT null or undefined, and then only add separator if it's not last item

illiarerun (changes requested):

illiarerun (approved):

PR #1724, opened 2022-10-25
Develop

AndreyChernysh:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
sh1gatsu, line 8: don't use names like `result`, use more meaningful names for your variables

sh1gatsu (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinString = '';
```
```suggestion
    let joinedString = '';
```

denys-danyliuk (approved):

PR #1734, opened 2022-10-26
Join method is ready

Sergey-Kovyazin:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
+      if (separator === undefined) {
+        newArray[i] = ',';
+      }
+    }
+
+    for (const element of newArray) {
+      result += element;
+    }
+
+    return result;
```
tetiana-ivanchenko, line 34: do you really need 4 loops to solve the task? actually you can do it using just 1 loop

tetiana-ivanchenko (changes requested):

Sergey-Kovyazin:
Like this?)

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        result += this[i];
+      }
+
+      if (i !== this.length - 1) {
+        if (separator === undefined) {
+          result += ',';
```
nazarmatsevych, line 17: You can remove this check if you set ',' as the default value for separator

nazarmatsevych (approved):
Yep, this solution is much better 👍 
Check comment pls

src/arrayMethodJoin.js
Sergey-Kovyazin: Thanks! Sometimes I make a much more difficult solution, that I need :)

PR #1737, opened 2022-10-26
add solution

AlexTsimba:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let joined = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
```
illiarerun, line 14: do not mutate an array, try to invert your if statement add add element here if it's NOT null or undefined

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+
+      if (i !== this.length - 1) {
+        joined += `${this[i]}${separator}`;
+      } else {
+        joined += `${this[i]}`;
+      }
```
illiarerun, line 20: and then only add separator here, like:
```suggestion
      if (i !== this.length - 1) {
        joined += separator;
      }
```

illiarerun (changes requested):

illiarerun (approved):

PR #1740, opened 2022-10-26
MethodJoin solution

shtht24:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        joinedString += '';
+      } else {
+        joinedString += `${this[i]}`;
+      };
+
+      if (i < this.length - 1) {
+        joinedString += `${separator}`;
```
sh1gatsu, line 18: same here (`'${}'` - delete it ;3)

sh1gatsu (changes requested):
Overall GJ, needs little fixes 👍

src/arrayMethodJoin.js
shtht24: ty for review, i've already fixed it ^_^

sh1gatsu (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        joinedString += '';
+      } else {
+        joinedString += this[i];
+      };
```
illiarerun, line 15: do you really need to add blank string? try to invert your if statement and then you wouldn't have to use else
```suggestion
      if (inverted statement) {
        joinedString += this[i];
      }
```

illiarerun (changes requested):

src/arrayMethodJoin.js
shtht24: yep, ty. that's much much better 👍

illiarerun (approved):

PR #1743, opened 2022-10-26
solution join

oksenyuk-valentyna:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
```
Oleksii-Siviolov, line 8: This is not the best variable name.
```suggestion
    let elementsString = '';
```

Oleksii-Siviolov (approved):
Good job in common, but take care of our mentors' nerves and use more meaningful variable names ;)

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let elementsString = '';
```
```suggestion
    let joinedString = '';
```

src/arrayMethodJoin.js
denys-danyliuk: this one too is not the best one ;)

denys-danyliuk (approved):

PR #1753, opened 2022-10-27
Solution

Bohdan-Kovalenko1992:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
```suggestion
    let joinedString = '';
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+    const lastElement = this[this.length - 1];
+
+    for (let i = 0; i < this.length; i++) {
+      if (this.length === 1) {
+        result += this[i];
+        break;
+      }
```
denys-danyliuk, line 15: why do you need this check?
if this.length === 1 it will work even without this check
also if this[i] is will be 'null' or 'undefined' this is check anyway will put it into result string

denys-danyliuk (changes requested):

Olexandr6 (approved):

PR #1757, opened 2022-10-30
Solution

natalia-maryshko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+      switch (true) {
+        case i === this.length - 1 && this[i] === undefined:
+          break;
+
+        case i === this.length - 1:
+          result += this[i];
+          break;
+
+        case separator === null:
+          result += this[i] + 'null';
```
polosanya, line 20: Why should you add word 'null' to your string?
Btw, you have default value for your separator, so it can't be null

polosanya (changes requested):
Good job! But do you really need to use switch(true)...case for this task? 
I think it can be solved easier. Please check my comments

src/arrayMethodJoin.js
natalia-maryshko: thank you!

polosanya (approved):
Well done!

PR #1759, opened 2022-10-30
first commit

haduigon:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,39 @@
...
+
+    if (typeof (separator) === 'object') {
+      result = copy.slice(15);
+    }
+
+    if (separator === null) {
+      result = copy.slice(4, -4);
+    }
+
+    return result;
```
tetiana-ivanchenko, line 39: you have a very confusing algorithm, all you need is two checks:
1. check if the element is not equal to null and undefined, add it to the resulting string
2. after that, check if the elementr is not the last one - add a separator to the resulting string

that's all)

tetiana-ivanchenko (changes requested):

PR #1763, opened 2022-11-04
test Ok

diolna:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let resultString = '';
+    let sepInsert = separator;
+
+    if (separator === undefined) {
+      sepInsert = ',';
+    }
```
nazarmatsevych, line 13: You can set ',' as a default value for separator

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      let element = this[i];
+
+      if (element === undefined || element == null) {
+        element = '';
+      }
```
nazarmatsevych, line 20:
```suggestion
      if (element !== null && element !== undefined) {
         resultString += this[i];
      }
```

We can improve the condition a bit

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
...
+
+      if (i !== this.length - 1) {
+        resultString += `${element}${sepInsert}`;
+      } else {
+        resultString += `${element}`;
+      }
```
nazarmatsevych, line 26:
```suggestion
      if (i !== this.length - 1) {
        resultString += separator;
      }
```

You can avoid else statement if you change the logic a little

nazarmatsevych (approved):
Good work 🔥 
Your solution works properly, but i have some improvements,
check comments

PR #1767, opened 2022-11-06
Solution

vitalii-borzov:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,41 @@
...
+      sep = 'null';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1) {
+        if (this[i] === undefined) {
+          result = result + '';
+
+          return result;
+        }
```
TarasHoliuk, line 26: this if statement will be redundant if you change logic a bit: add separator only if it's not the last element

TarasHoliuk: and in this case, `return` inside the loop will be redundant

TarasHoliuk (changes requested):

src/arrayMethodJoin.js
vitalii-borzov: I can't figure out what to do (
Tell me, please.

tetiana-ivanchenko: why do you need this statement, if you add just empty string it's mean that you didn't add anything

tetiana-ivanchenko: you need just two checks:
1. if current element isn't null and isn't undefined, you shoud add it to resulted string
2. if current element is not last, add separator to your resulted string

tetiana-ivanchenko (changes requested):

PR #1769, opened 2022-11-06
First commit Mettod join

andry5bb:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,41 @@
...
+    if (this.length === 0) {
+      return result;
+    }
+
+    if (separator === null) {
+      sep = 'null';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1) {
```
TarasHoliuk, line 21: write the if statement in the opposite way:
add a separator if it's not the last element

In such a case nested if will be unnecessary

TarasHoliuk (changes requested):

PR #1774, opened 2022-11-11
method join was implemented

StetsenkoViktoriia:


src/arrayMethodJoin.js
```diff
@@ -1,11 +1,24 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null) {
+        if (this[i] !== undefined) {
+          result += this[i];
+        }
+      }
```
liliia-livandovskaya, line 12:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        result += this[i];
      }
```

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,24 @@
...
+
+      if (i !== this.length - 1) {
+        if (separator !== undefined) {
+          result += separator;
+        }
+      }
```
liliia-livandovskaya, line 18:
```suggestion
      if (i !== this.length - 1 && separator !== undefined) {
        result += separator;
      }
```
avoid nesting, use logical operators

liliia-livandovskaya (changes requested):

polosanya (approved):
Good job!

PR #1777, opened 2022-11-13
solution

olenitut:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,24 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = String(this[0]);
+
+    if (this[0] === null || this[0] === undefined) {
+      result = '';
+    }
```
Bohdan-Maksymiuk, line 12:
```suggestion
    let result = '';
```
you have the same logic into the loop
so lets keep our code DRY
remove it here and iterate from the first element

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #1778, opened 2022-11-14
Solution

anton-sosonskyi:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,30 @@
...
+    let resultStr = '';
+    let joincharacter = separator;
+
+    if (separator === '') {
+      joincharacter = '';
+    } else if (separator === null) {
+      joincharacter = 'null';
+    } else if (separator === undefined) {
+      joincharacter = ',';
+    }
```
Bohdan-Maksymiuk, line 18: avoid such nesting `else if` conditions

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,30 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1 && this[i] === undefined) {
+        continue;
+      } else if (i === this.length - 1) {
+        resultStr += this[i];
+      } else if (this[i] === null || this[i] === undefined) {
+        resultStr += '' + joincharacter;
+      } else {
+        resultStr += this[i] + joincharacter;
+      }
```
Bohdan-Maksymiuk, line 29: avoid such nesting `else if` conditions

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #1779, opened 2022-11-15
WWWWW

yurii-filiushkin:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,46 @@
...
+
+      default:
+        newString = copyArray[0].toString();
+
+        for (let index = 1; index < copyArray.length; index++) {
+          newString += newSeparator.toString() + copyArray[index];
+        }
+    }
+
+    return newString;
```
tetiana-ivanchenko, line 46: you have very confuse algorithm. actualy in the loop, you need just two checks:
1. if the current element isn't null and isn't undefined - add it to the resulted string
2. if the current is't last - add separator to the resulted string

that's all

tetiana-ivanchenko (changes requested):

src/arrayMethodJoin.js
yurii-filiushkin: Ok)))

TarasHoliuk (approved):

PR #1780, opened 2022-11-15
Solution

Nazar-Brunarskyi:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const spr = `${separator}`;
+    let joinedString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let e = this[i];
```
Oleksandr-Filo, line 12: you should use more understandable names for variables (spr, e)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+
+      if (e === null || e === undefined) {
+        e = '';
+      }
+
+      joinedString += spr + e;
+    }
+
+    return joinedString.slice(spr.length);
```
Oleksandr-Filo, line 21: is it allowed to use any method in this task?

Oleksandr-Filo (approved):

src/arrayMethodJoin.js
Nazar-Brunarskyi: my logic said that's a method of String so I used it)

ivan-kadykalo (approved):
🤘

PR #1782, opened 2022-11-16
Solution

OksanaBaloh:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const lastElement = this.length - 1;
+
+    for (let i = 0; i < lastElement; i++) {
+      if (this[i] === null || this[i] === undefined) {
```
Viktor-Kost, line 12: Here you can simplify your condition just using `!this[i]`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+        result += `${separator}`;
+        continue;
+      }
+
+      result += this[i] + `${separator}`;
+    }
+
+    if (this[lastElement] !== null && this[lastElement] !== undefined) {
```
Viktor-Kost, line 20: Similar to previous comment)

Viktor-Kost (changes requested):
Nice) 
Make some improvements)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const lastElement = this.length - 1;
```
loralevitska, line 9: [useful link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at)

```suggestion
    const lastElement = this.at(-1)
```

loralevitska (approved):
great!

src/arrayMethodJoin.js
OksanaBaloh: But element "0" of array also `false`, and I will have a mistakes, if I use `!this[i]`.

src/arrayMethodJoin.js
OksanaBaloh: I tried, but VSCode says "TypeError:` this.at` is not a function". In Slack, another mentor said that the reason of this error is that  the tests don't know about `at()` method.

PR #1783, opened 2022-11-16
solution

panianchuk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
```
Viktor-Kost, line 11: Here you can simplify your code using !this[i]

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+        this[i] = '';
+      }
+
+      if (i === 0) {
```
Viktor-Kost, line 15: It seems like you don't need to use this check with i === 0

Viktor-Kost (changes requested):

src/arrayMethodJoin.js
Viktor-Kost: I meant` if (!this[i]) {}`

Viktor-Kost (approved):
Nice!

PR #1788, opened 2022-11-20
Solution

IVxVI:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+        res += value;
+      }
+
+      if (this.indexOf(value) < this.length - 1) {
+        res += separator;
+      }
+    }
+
+    return res;
   };
```
SanyaBratashchuk, line 21: use common `for` loop to avoid indexOf which increase complexity of your alghorithm

SanyaBratashchuk (changes requested):

ivanchuk-yurii (approved):

PR #1790, opened 2022-11-22
js_array-method-join was added

vadiimvooo:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        this[i] = '';
+      }
+
+      if (i !== this.length - 1) {
+        stringJoin += String(this[i]) + separator;
+      } else {
+        stringJoin += String(this[i]);
+      }
```
official-artem, line 19:
```suggestion
      if (this[i] !== undefined || this[i] !== null) {
         stringJoin += String(this[i]);
      }

      if (i !== this.length - 1) {
        stringJoin += separator;
```

official-artem (approved):

src/arrayMethodJoin.js
Ihor-Karpyn: Хороща порада

Ihor-Karpyn (approved):

PR #1791, opened 2022-11-22
Solution

RomaSheva1987:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringWithoutLastElement = '';
+    const lastElement = this[this.length - 1];
+    let x = separator;
+
+    if (x === null) {
```
vlad-demchuk, line 12: x can't be null because you set a default value for separator

vlad-demchuk: `x` is redundant, use `separator`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringWithoutLastElement = '';
```
vlad-demchuk, line 8: redundant variable

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+    const lastElement = this[this.length - 1];
+    let x = separator;
+
+    if (x === null) {
+      x = 'null';
+    }
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] === null || this[i] === undefined) {
```
vlad-demchuk, line 17: just check if this[i] exists and check if it's not the last element
and if it's not add separator, otherwise return result string

vlad-demchuk (changes requested):

vlad-demchuk (approved):

PR #1793, opened 2022-11-22
Solution

UmizDemud:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        result += this[i];
+      }
+      result += sep;
+    }
+
+    // adding the last element
+    if (this[this.length - 1] !== null
+        && this[this.length - 1] !== undefined) {
```
mgrinko, line 28: the condition for the element check is duplicated 
please avoid the duplication

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    let sep;
+
+    // in case the passed separator is null
+    if (separator === null) {
+      sep = 'null';
+    } else {
+      sep = separator;
+    }
```
mgrinko, line 16: ?:

mgrinko (changes requested):

src/arrayMethodJoin.js
UmizDemud:
```js
test(`Should work with null`, () => { // here is the test 
  expect(source.join2(null))
    .toBe('0null1null2null3');
});

Comment (meforoma) [reply src/arrayMethodJoin.js:16]: i guess with this `?:` Misha meant the ternary
```suggestion
    separator === null
      ? sep = 'null';
      : sep = separator;
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+
+    // joining the array excluding the last element
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        result += this[i];
+      }
+      result += sep;
```
meforoma, line 23: ternary

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+    // in case the passed separator is null
+    if (separator === null) {
+      sep = 'null';
+    } else {
+      sep = separator;
+    }
+
+    // joining the array excluding the last element
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
```
meforoma, line 20: i suppose a generic check would do the job?
```suggestion
      if (this[i]) { ...
```

meforoma (changes requested):
see the comments

src/arrayMethodJoin.js
UmizDemud: 🤦‍♂️ ..

src/arrayMethodJoin.js
UmizDemud: I tried this, the case where an element is 0 is the problem.

src/arrayMethodJoin.js
UmizDemud: Correcting

naziamoff (approved):
Well done!

PR #1794, opened 2022-11-22
Task was done

Sergey-Mironenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+        step = '';
+      } else {
+        step = this[i];
+      }
+
+      if (i === this.length - 1) {
+        result += `${step}`;
+      } else {
+        result += `${step}` + `${separator}`;
+      }
```
tetiana-ivanchenko, line 22:
```suggestion
      if (this[i] !== null || this[i] !== undefined) {
        result += this[i];
      }

      if (i === this.length - 1) {
        result += separator;
      }
```

tetiana-ivanchenko: also why did you use interpolation here  `${step}`, it's redundant?

tetiana-ivanchenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
tetiana-ivanchenko, line 8:
```suggestion
    let joinedString = '';
```

tetiana-ivanchenko (changes requested):
please, pay attention to whether all the comments that mentors left is fixed
https://github.com/mate-academy/js_array-method-join/pull/1794/files#r1030208079

tetiana-ivanchenko (approved):

PR #1795, opened 2022-11-23
Method join was implemented

serge-kontakt:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = ``;
```
```suggestion
    let result = '';
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
...
+      } else {
+        result += `${this[i]}`;
+      }
+
+      if (i < (this.length - 1)) {
+        if (separator === undefined) {
+          result += `,`;
+        } else {
+          result += `${separator}`;
+        }
```
liliia-livandovskaya, line 22:
```suggestion
      if (this[i] !== undefined & this[i] !== null) {
        result += this[i];
      }

      if (i !== this.length - 1) {
        result += separator;
      }
```

liliia-livandovskaya (changes requested):
Pay attention to the comments

src/arrayMethodJoin.js
serge-kontakt: Thank you for the recommendations. Unfortunately not all of them were totally correct in tests were separator is not specified or undefined. So I had to improve them.

liliia-livandovskaya (approved):

PR #1796, opened 2022-11-23
Solution

petrolozynskyi:


stanislav-topikha (changes requested):
try to reduce number of `if` statements

PR #1798, opened 2022-11-23
join method

mr-valentos:


GabrielMaksimovich (approved):
Nothing to add!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+    for (let i = 0; i < this.length; i++) {
+      let checkedElement = this[i];
+
+      if (checkedElement === null || checkedElement === undefined) {
+        checkedElement = '';
+      }
+
+      if (i === 0) {
+        joinString += checkedElement;
+      } else {
```
valerii-derkach, line 21: you can add this element at the beginning and start iterating from index 1, no need an extra check

valerii-derkach (changes requested):

valerii-derkach (approved):

PR #1801, opened 2022-11-24
solution

nikachu404:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+      } else {
+        result += this[i] + separatorString;
+      }
+    }
+
+    if (lastElement !== undefined) {
+      result += lastElement;
+    }
+
+    return result;
```
4ebupel, line 24: Maybe, it's better not to use 'result' as a name for variable
<sub>Just my opinion</sub>

4ebupel (approved):
Good job!

Ihor-Karpyn (approved):

PR #1803, opened 2022-11-24
Solution :')

CharAnagn:


IoghiStudio (approved):
You could simplify it a bit but if tests pass that's all good!

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,34 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let result = '';
+    let sep = separator;
+
+    if (separator === null) {
+      sep = 'null';
+    }
```
naziamoff, line 14: it should work the same way even without this check I guess

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,34 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== this[this.length - 1]) {
+        if (this[i] !== undefined && this[i] !== null) {
+          result += this[i] + sep;
+          continue;
+        }
```
naziamoff, line 21: nested if's are evil 👿 
Any ideas on how we can avoid them here?

naziamoff (changes requested):

src/arrayMethodJoin.js
CharAnagn: I was a bit frustrated and wanted it to work 😄 Now, with a clear mind I think I got it.

naziamoff (approved):
great!

PR #1805, opened 2022-11-24
solution

Lun4trick:


CharAnagn (approved):
My solution looks like a mess compare to yours, my brain got melted 🤣

Gj 🔥

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const checkSep = (separator === null) ? 'null' : separator;
+
+    for (let i = 0; i < this.length - 1; i++) {
```
meforoma, line 11: do you think it would be easier if you use a `for..of` loop?
```suggestion
    for (const element of this) { ...
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+      if (this[i] === null || this[i] === undefined) {
```
meforoma, line 12: not totally sure, but do you need an explicit check for `null` and `undefined` or could just go with a generic check `if (this[i])` that would result in false if this[i] is any equivalent of a `falsy` value

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+        result += '' + checkSep;
+      } else {
+        result += this[i] + checkSep;
+      }
+    }
+
+    return (this[this.length - 1] === undefined)
+      ? result : result + this[this.length - 1];
```
meforoma, line 20: multi-line and generic check?
```suggestion
    return (this[this.length - 1])
      ? result + this[this.length - 1]
      : result;
```

meforoma (changes requested):
see the comments - validate and apply if they're relevant

src/arrayMethodJoin.js
Lun4trick: i used this type of loop cause i need to run it once less than the actual length.
thats cause at the end of the code, it checks the last element of the array, and if its falsy its not gonna added to the result.

src/arrayMethodJoin.js
Lun4trick: yeah for first i wanted to do that way, but in this case its gonna ignore all false values like 0 or NaN, and it sould not to be ignored by the code.

src/arrayMethodJoin.js
Lun4trick: yeah i changed it, it works well, but what if the last element is 0 or NaN(the tests not checks this), thats gonna be not added at the end too right? Thank you for the suggestions! :)

PR #1807, opened 2022-11-25
solution

lefteris-L:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const lastElement = this[this.length - 1];
+    let string = '';
+
+    for (const element of this) {
+      const elem = element !== null && element !== undefined
```
UmizDemud, line 12: I think you can just write 
```js
const elem = element != null
  ? ...
  : ...;

File: src/arrayMethodJoin.js [hunk 2]
```diff
@@ -4,8 +4,21 @@
...
+        ? element
+        : '';
+
+      string += element !== lastElement
```
UmizDemud, line 16: This checks for the equality with lastElement. In case we have 
```js
['asd', '1', '2', 'asd', 'asd'].join2(',');
```

I believe the code would result 
```js
asd1,2,asdasd
```

Consider looping through indexes:
```js
for (let i = 0 ...

UmizDemud (changes requested):
Good approach. I have two comments, please check 'em out.

src/arrayMethodJoin.js
lefteris-L: Tried that. You need to explicitly check for undefined as well

src/arrayMethodJoin.js
lefteris-L: You're right, index would be a more reliable comparison

UmizDemud (approved):
Approved upon change

naziamoff (changes requested):

naziamoff (approved):

PR #1808, opened 2022-11-25
Solution

vitalii-borzov:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,41 @@
...
+
+      if (this[i] === null || this[i] === undefined) {
+        addToString = '' + sep;
+      } else {
+        addToString = this[i] + sep;
+      }
+      result = result + addToString;
+    }
+
+    return result;
```
tetiana-ivanchenko, line 41: you have so much conditions, but actually you need only two.
if (this[i] !== null || this[i] !== undefined)
add this[i] to the resulting string, than if your element is not last - add separator to the resulting string, that's all)

tetiana-ivanchenko (changes requested):

PR #1810, opened 2022-11-25
Solution

Deniskakaka:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,16 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    const separator2 = separator === undefined ? ',' : `${separator}`;
```
liliia-livandovskaya, line 9:
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,16 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
liliia-livandovskaya, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,16 @@
...
+    let result = '';
+    const separator2 = separator === undefined ? ',' : `${separator}`;
+
+    for (let i = 0; i < this.length; i++) {
+      result += this[i] !== null && this[i] !== undefined
+        ? this[i].toString() + separator2
+        : '' + separator2;
+    }
+
+    return separator2.length < 1 ? result : result.slice(0, -separator2.length);
```
liliia-livandovskaya, line 17: try to rewrite this line, it's hard to read it

liliia-livandovskaya (changes requested):

Virass (changes requested):
Almost done, but you did not take into account all the comments above. Set default value for function argument

src/arrayMethodJoin.js
Virass: And make changes for this comment

Virass (changes requested):
Make changes according to all the comments above that Lilya said

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      result += this[i] !== null && this[i] !== undefined
+        ? this[i].toString() + separator
+        : '' + separator;
+    }
+
+    switch (true) {
```
Viktor-Kost, line 16: This switch statement with one case doesn't make sense
You can just use early return

```
if(`${separator}`.length >= 1) {
  return result.slice(0, -`${separator}`.length);
}

return result;
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      result += this[i] !== null && this[i] !== undefined
```
Viktor-Kost, line 11: It is not readable. Make it simple. Just if statement

Viktor-Kost (changes requested):

src/arrayMethodJoin.js
Deniskakaka: тут только два исключения, как их ещё проще написать можно?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const sep = i < this.length - 1 ? separator : '';
```
witflash, line 11: You can avoid this constant if you will use two sequential `if` blocks.
Smth like this:
```
  // do smth
}

if (I < this.length) {
  // do smth
}
```

You can try it on your own.

witflash (approved):
Much better! Good job!

PR #1816, opened 2022-11-26
solution

mmahfuzi:


IoghiStudio (commented):
You could simplify it a little.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,37 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let sep = separator;
+
+    if (this.length === 0) {
+      return '';
+    }
```
Mohameddahi, line 13: there is no need for these lines because in line 27 you have assigned the first element  of the array to the result and  when the array is empty , it will return empty string.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,37 @@
...
+      return String(this);
+    }
+
+    if (sep === null) {
+      sep = 'null';
+    }
+
+    if (this[0] === null || this[0] === undefined) {
+      this[0] = '';
+    }
```
Mohameddahi, line 25: and you can put all these conditions inside the loop.

Mohameddahi (changes requested):
in general, it is good. But to make it better, consider my comments.

naziamoff (approved):
good job!

PR #1826, opened 2022-11-28
Solution

kattentionn:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
```
official-artem, line 8: Mentors don't like a 'result' word. You should use “changedString” for example or something like this.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+    const lastInArray = this[this.length - 1];
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        resultString += `${separator}`;
+      } else {
+        resultString += this[i] + `${separator}`;
+      }
```
official-artem, line 16:
```suggestion
      }
      
      resultString += this[i] + `${separator}`;
```

official-artem (approved):

src/arrayMethodJoin.js
kattentionn: thank you, I will change it)

src/arrayMethodJoin.js
kattentionn: It better works with 'else', so I will leave it

official-artem: > It better works with 'else', so I will leave it

Yeah, my bad, because in this case you need to use break in 'if' condition. You can leave it.

valerii-derkach (approved):

PR #1830, opened 2022-11-28
solution

Alireza-Mohammadhossein:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,26 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let newSeparator = separator;
+
+    if (separator === undefined) {
```
UmizDemud, line 10: You can use ternary operator to make this if unnecessary.

```js
const newSeparator = separator === undefined ? ',' : separator;
```

This would assign separator if it exists, if it doesn't

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,26 @@
...
+      newSeparator = ',';
+    }
+
+    let list = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
```
UmizDemud, line 17: You can use:
```js
 if (this[i] != null) {
 ```
 The logic here is that
 ```js
 (null == undefined) === true //  and shortens the code a little

**Review by UmizDemud — CHANGES_REQUESTED** [review]
Code seems correct, could found two improvements. Please take a look and I'll review/approve

File: src/arrayMethodJoin.js (hunk 2 above)
Comment (Alireza-Mohammadhossein) [reply src/arrayMethodJoin.js:17]: thank you, Umut, but when I use this[i] != null it returns an error to me, saying that I should use  !== null

File: src/arrayMethodJoin.js (hunk 1 above)
Comment (Alireza-Mohammadhossein) [reply src/arrayMethodJoin.js:10]: Good idea

File: src/arrayMethodJoin.js [hunk 3]
```diff
@@ -5,8 +5,22 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const newSeparator = separator === undefined ? ',' : separator;
```
naziamoff, line 8: I understand the temptation to provide one line solution but it is better to keep it multi-lined with ternary))
And what's more, you can use shortcut for checking whether the separator is defined

```suggestion
    const newSeparator = separator 
      ? ',' 
      : separator;
```

naziamoff (approved):
looks good to me)
But you should keep ternary multi-lined

PR #1831, opened 2022-11-29
Solution

4ebupel:
Brain is melting

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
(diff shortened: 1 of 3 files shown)


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
official-artem, line 8: Don't use the name 'result' for a variable. You should rename this variable like 'changedString 'or smt like this.

official-artem (changes requested):

src/arrayMethodJoin.js
4ebupel: Fixed, I guess

official-artem (approved):

Ihor-Karpyn (approved):

PR #1835, opened 2022-11-29
Solution

Mohameddahi:


Alireza-Mohammadhossein (commented):
Everything is perfect

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
...
+        this[i] = '';
+      }
+
+      if (separat === undefined) {
+        separat = ',';
+      }
+
+      if (separat === null) {
+        separat = 'null';
+      }
```
naziamoff, line 22: do you really need this one?

naziamoff (changes requested):

src/arrayMethodJoin.js
naziamoff: still unanswered

naziamoff (approved):

PR #1838, opened 2022-11-29
Solution

Grayfox147:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,35 @@
...
+      this[0] = '';
+    }
+
+    let result = this[0];
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
+      result = result + sep + this[i];
```
ysrckr, line 32: You can write result += sep + this[i]

ysrckr (approved):
Other than the comment, it looks ok.

src/arrayMethodJoin.js
Grayfox147: Done! thank you :D

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,35 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
+
+    if (this.length === 1) {
+      return String(this);
+    }
```
naziamoff, line 14: do you really need those checks?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,35 @@
...
+
+    let sep = separator;
+
+    if (sep === null) {
+      sep = 'null';
+    }
+
+    if (this[0] === null || this[0] === undefined) {
+      this[0] = '';
+    }
```
naziamoff, line 24: you repeat the same logic inside the loop below, can we make it only in one place?

naziamoff (changes requested):

naziamoff (approved):
🔥

PR #1848, opened 2022-11-29
Solution for array method join

katerinalex:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let myString = '';
+
+    if (this.length < 1) {
+      return myString;
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      // myString = myString + separator;
```
polosanya, line 15: Don't forget to delete comments

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+
+      if (this[i] === 'null') {
+        myString = myString + 'null';
+      } else if (this[i] === false && typeof (this[i]) === 'boolean') {
+        myString = myString + 'false';
+      } else if (this[i] !== null && this[i] !== undefined && this[i] !== '') {
```
polosanya, line 21: Avoid using else...if. For now your code looks complicated. 
Try using switch...case and/or ternary operator instead.

polosanya (changes requested):
Good job! Please check the comments to improve your code a bit

Bohdan-Maksymiuk (approved):

PR #1850, opened 2022-11-29
Method join

DiDemkiv:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        newJoin += this[i] + separatorToString;
+      } else {
+        newJoin += separatorToString;
+      }
+    }
+
+    if (lastSymbol !== null && lastSymbol !== undefined) {
+      newJoin += lastSymbol;
```
GabrielMaksimovich, line 21: I think it should work without this one "lastSymbol !== null &&" it works just with lastSymbol !== undefined

GabrielMaksimovich (changes requested):

src/arrayMethodJoin.js
DiDemkiv: you're right, fixed :)

valerii-derkach (approved):

PR #1874, opened 2022-11-30
implemented join method

TheOneThatKnowsNothing:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = (this[0] === null || this[0] === undefined)
+      ? ''
+      : String(this[0]);
+    const elementSerarator = separator === null
```
Oleksandr-Filo, line 11:
```suggestion
    const elementSeparator = separator === null
```

Oleksandr-Filo (approved):
looks good

Ihor-Karpyn (approved):

src/arrayMethodJoin.js
Ihor-Karpyn: Let's fix this var

PR #1877, opened 2022-12-01
Solution

batstolya:


viktoriia-kravchenko (approved):
🔥

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const separatorToString = separator + '';
```
solaryasha, line 8:
```suggestion
    const separatorToString = String(separator);
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+      }
+    }
+
+    const lengthSeparator = separatorToString.length;
+
+    if (lengthSeparator === 0) {
+      return resultString;
+    }
+
+    const stringWithoutLastElement = resultString.slice(0, -lengthSeparator);
```
solaryasha, line 27: Could we just check it in the loop ?

solaryasha (changes requested):

solaryasha (changes requested):
Comments still not fixed.

solaryasha (changes requested):
Push your code to the github

src/arrayMethodJoin.js
batstolya: > Push your code to the github

@solaryasha  To be honest, I have no idea how to put the last check in the loop. I spent about an hour and it didn't work out for me) So I made another solution

solaryasha (approved):
👍

PR #1879, opened 2022-12-01
Added applyCustomJoin function

viktoriia-kravchenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+    }
+
+    if (this[0] === null) {
+      customJoin += '';
+    } else {
+      customJoin += this[0];
+    }
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null || this[i] === ' ') {
```
batstolya, line 21: do we need this test?  `this[i] === '  ' ` code works without it.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+        customJoin += String(separator) + '';
```
batstolya, line 22:
```suggestion
        customJoin += String(separator);
```
customJoin and String(separator) == string anyway

batstolya:
the rest looks good to me 🔥

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let customJoin = '';
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    if (this[0] === null) {
```
solaryasha, line 14: This also true for `undefined` value

solaryasha (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let customJoin = '';
+
+    if (this.length === 0) {
```
ivan-kadykalo, line 10:
```suggestion
    if (!this.length) {
```

PR #1881, opened 2022-12-01
solution

vladimir-arkanov:


Anastasia4sik (approved):
Looks good!
you could replace if...else with one if but it is not so important

ivan-kadykalo (approved):
Good job 🔥

PR #1889, opened 2022-12-04
 js_array-method-join

rafmamedov:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (const value of this) {
+      if (this.indexOf(value) === this.length - 1) {
```
ivanchuk-yurii, line 11: why do you need this? just use for (let i = 0; ...)

ivanchuk-yurii (changes requested):

src/arrayMethodJoin.js
rafmamedov: @ivanchuk-yurii I changed, but I didn't understand how this simplifies the solution. Can you give me more details please, what I was supposed to change in the code?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
```
ivanchuk-yurii, line 10: `const value = this[i];`

ivanchuk-yurii (approved):

PR #1892, opened 2022-12-06
Initial commit

yaroslav-furhala:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const element = this[i];
+
+      string += (element === null || element === undefined)
+        ? ''
```
vlad-demchuk, line 14: you don't need this case. just you `if...else` here

vlad-demchuk (changes requested):

stanislav-topikha (approved):

PR #1899, opened 2022-12-10
Solution method join

Sasha-Green:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+    let strArray = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let element = this[i];
+
+      if (this[i] === undefined || this[i] === null) {
+        element = '';
+      }
+
+      strArray += element;
```
TarasHoliuk, line 17: actually, adding an empty string does nothing, so change the condition to `this[i] !== undefined || this[i] !== null` and move strArr += element to `if` block

TarasHoliuk (changes requested):

loralevitska (approved):
good job!

PR #1909, opened 2022-12-18
Solution method join

anna-akhalmosulishvili:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,39 @@
...
+          joinArr[i] = joinElement;
+        } else {
+          joinArr[i] = '';
+        }
+      }
+    }
+
+    for (const i of joinArr) {
+      joinStr += i;
+    }
```
SanyaBratashchuk, line 38: simplify your code all what you need is set `','` as default value for `separator` parametr then iterate through string and check on each iteration whether the current character is neither `null ` nor `undefined` if no, concate current char with result string, then check if current index is not last, and concate separator with result string. That's it

SanyaBratashchuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinStr = '';
+
+    if (this.length === 0) {
+      return joinStr;
+    }
```
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if ((this[i] !== null) && (this[i] !== undefined)) {
```
SanyaBratashchuk, line 15:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
```

SanyaBratashchuk (approved):

PR #1910, opened 2022-12-18
Solution

yana-kotsulym:


SanyaBratashchuk (changes requested):
simplify your code all what you need is set ',' as default value for separator parametr then iterate through string and check on each iteration whether the current character is neither null nor undefined if no, concate current char with result string, then check if current index is not last, and concate separator with result string. That's it

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const array = this;
```
DarkMistyRoom, line 8: Why do you need this copy?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+    let resultString = '';
+
+    for (const item of array) {
+      if (item === null || item === undefined) { } else {
+        resultString += item;
+      }
```
DarkMistyRoom, line 14:
```suggestion
      if (item !== null && item !== undefined) {
        resultString += item;
      }
```

src/arrayMethodJoin.js
DarkMistyRoom: you can work with `this`

src/arrayMethodJoin.js
DarkMistyRoom: It is bad practice

DarkMistyRoom (approved):

PR #1914, opened 2022-12-21
joinFunction

Oleksandr-Maslo:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+    let copySeparator = separator;
+
+    if (copySeparator === null) {
+      copySeparator = 'null';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
```
nazarmatsevych, line 18:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        result += this[i];
      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+
+      if (i < this.length - 1) {
+        result += this[i] + copySeparator;
+      } else {
+        result += this[i];
+      }
```
nazarmatsevych, line 24:
```suggestion
      if (i < this.length - 1) {
        result += separator;
      }
```

nazarmatsevych (changes requested):
Some improvements in comments

nazarmatsevych (approved):

PR #1915, opened 2022-12-21
finished

barantarasNEW:


witflash (approved):
Good job!
You can do this without `switch/case`:
```
  result += this[I];
}

if (i < this.length) {
  result += separator;
}

return result;
```

Something like this

Bohdan-Maksymiuk (approved):

PR #1916, opened 2022-12-22
custom join method completed

MykolaKaradzha:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let newSeparator = separator;
+    const newArray = [];
+    let string = '';
+
+    if (!separator) {
+      newSeparator = `${separator}`;
+    }
```
Viktor-Kost, line 14: Why do you need this check? You have already set a default value for your separator parameter above `separator = ','`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+      }
+    }
+
+    for (let i = 0; i < newArray.length; i++) {
+      if (i === newArray.length - 1) {
+        string += newArray[i];
+      } else {
+        string += newArray[i] + newSeparator;
+      }
+    }
```
Viktor-Kost, line 30: We should not use a loop twice with a buffer variable like new Array. 
Just inside your first loop make checks

```
if () {
  string += this[i];
}
if (...) {
  string += separator;
}
```

This is all that you need)

Viktor-Kost (changes requested):

src/arrayMethodJoin.js
MykolaKaradzha: Well, I`ve done it at first, but  if i don`t do this,  test with null will ignore null as separator. 
(test(`Should work with null`, () => {
  expect(source.join2(null))
    .toBe('0null1null2null3');
});)

src/arrayMethodJoin.js
MykolaKaradzha: Well, newArray is needed for this test:
 test(`Should work when array has null and undefined`, () => {
  expect([false, null, 1, undefined, 0, NaN, ''].join2(','))
    .toBe('false,,1,,0,NaN,');
});

It filters undifned and null from array.

MykolaKaradzha: Got you, but I can`t resolve problem with null and undefined. Here is my code:
image
Unfortunately, I can`t pass tests, because this condition is not working for some reason.

Viktor-Kost: So i meant just use your conditions at if statement
```
  string += this[i];
}
if (i === newArray.length - 1) {
  string += separator;
}
```

Viktor-Kost (changes requested):
Hello!) 
Add an answer to your question. 
If you have any questions - please, ask at fe_chat

AlisaVlasova (approved):

PR #1917, opened 2022-12-22
solution

drhtka:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let tmpString = '';
+    const lastKey = this.length - 1;
+
+    for (const key of this) {
+      if (key === null || key === undefined) {
+        tmpString += '';
+      } else {
+        tmpString += key;
+      }
```
tetiana-ivanchenko, line 16:
```suggestion
      if (key !== null || key !== undefined) {
        tmpString += key;
      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+
+      if (key !== this[lastKey]) {
```
tetiana-ivanchenko, line 18: what if you will have this array:
[1, 2, 3, 4, 2]
how this condition will works?
you will receive string: "1.23.4.2"
original join doesn't works same

tetiana-ivanchenko (changes requested):

src/arrayMethodJoin.js
TarasHoliuk: fix it
as Tetiana said it will work in the wrong way
use common loop `for` instead of `for of` and check index

TarasHoliuk (changes requested):

SanyaBratashchuk (approved):

PR #1922, opened 2022-12-26
add solution

Anastasiia-Khudych:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,47 @@
...
+
+    if (this.length === 1) {
+      string = `${this[0]}`;
+
+      return string;
+    }
+
+    if (currentSeparator === undefined) {
+      currentSeparator = ',';
+    }
```
Kravchenko71213, line 27: I think it's better to write it like this

Kravchenko71213: function(separator = ',')

Kravchenko71213 (approved):

src/arrayMethodJoin.js
Anastasiia-Khudych: Thanks, you are right)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,43 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (i === 0 && this[i] === null) {
+        string = '';
+        continue;
+      }
+
+      if (i === 0 && this[i] !== null) {
+        string = `${this[i]}`;
+        continue;
+      }
```
ArturErmolenko, line 33:
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,43 @@
...
+
+      if (this[i] === null || this[i] === undefined) {
+        string = `${string}${separator}`;
+        continue;
+      }
+
+      string = `${string}${separator}${this[i]}`;
+    }
```
ArturErmolenko, line 41:
```suggestion
      if (this[i] !== null || this[i] !== undefined) {
        string += this[i];
      }

      if (i !== this.length - 1) {
         string += separator;
       }
    }
```

ArturErmolenko (approved):

PR #1926, opened 2022-12-28
Solution

ruslan-yarosh:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,36 @@
...
+    for (let i = 0; i < this.length; i += 1) {
+      if (this[i] === undefined || this[i] === null) {
+        result += '';
+      } else {
+        result += this[i];
+      }
+
+      if (i === this.length - 1) {
+        return result;
+      }
```
ViktorHavryliuk, line 19: You don't need this check, remove it please

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,36 @@
...
+      switch (separator) {
+        case undefined:
+          result += ',';
+
+          break;
+
+        case null:
+          result += 'null';
+
+          break;
```
ViktorHavryliuk, line 30: Instead of checks for null and undefined better just set a default value of ',' for separator argument

ViktorHavryliuk (changes requested):

src/arrayMethodJoin.js
ruslan-yarosh: Thank you! I will keep in mind

loralevitska (approved):
good job!

PR #1927, opened 2022-12-28
Solution

Ilya-Osoben:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+      if (i < this.length - 1) {
+        string += separator;
+      }
+    }
+
+    const lastElement = this[this.length - 1];
+
+    if (lastElement === undefined || isNaN(lastElement)) {
+      string = string.slice(0, -1);
+    }
```
TarasHoliuk, line 27: why do you need it? if the last element is, for example `'a'` you will remove it?
image

TarasHoliuk (changes requested):

PR #1929, opened 2022-12-28
Solution

anna-inozemets:
Custom method join was realised

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let strFromArr = '';
+
+    const sep = `${separator}`;
+
+    for (const item of this) {
+      if (item === null || item === undefined) {
+        strFromArr += `${sep}`;
+      } else {
+        strFromArr += item + `${sep}`;
+      }
+    }
+
+    return strFromArr.slice(0, strFromArr.length - sep.length);
   };
 }
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -463,13 +463,248 @@
       }
     },
     "@mate-academy/scripts": {
-      "version": "0.3.5",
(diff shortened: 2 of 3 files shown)


andrieieva-yuliia (approved):
Nice code 👩‍💻

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let strFromArr = '';
+
+    const sep = `${separator}`;
```
Anna-Tkachenko, line 10: do you need this?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let strFromArr = '';
```
Anna-Tkachenko, line 8: will be better to create understandable variable name instead of using name of all used data types here

Anna-Tkachenko (changes requested):

src/arrayMethodJoin.js
anna-inozemets: As I use slice, when I don't use interpolation I have an error 'TypeError: Cannot read property 'length' of null',  so I created an extra variable. But, As i recognised it can be too complivated, so I try to create a another way to solve this task

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,38 @@
...
+
+    return joinedArray.slice(0, joinedArray.length - `${separator}`.length);
+  };
+} */
+
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedArray = '';
```
Anna-Tkachenko, line 25: will be better not to use data type names for variable names

Anna-Tkachenko (approved):

PR #1930, opened 2022-12-28
Solution

Kardash-Oleksandr:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,37 @@
...
+      checkedSeparator = 'null';
+    } else {
+      checkedSeparator = separator;
+    }
+
+    for (let element = 0; element < this.length; element++) {
+      if (this[element] === undefined) {
+        if (element !== this.length - 1) {
+          joinedArray += checkedSeparator;
+        }
```
Anastasiia-Khudych, line 21: The 4th item in the checklist (avoid using nested if statements).

Anastasiia-Khudych (changes requested):
Well done!

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,29 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedArray = '';
+    let checkedSeparator;
+
+    if (separator === null) {
+      checkedSeparator = 'null';
+    } else {
+      checkedSeparator = separator;
+    }
```
ArturErmolenko, line 12: I believe you don't need it

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,29 @@
...
+      }
+
+      if (element === this.length - 1) {
+        joinedArray += this[element];
+        continue;
+      }
+      joinedArray += this[element] + checkedSeparator;
+    }
+
+    return joinedArray;
```
ArturErmolenko, line 26: if you go backwards your code will be more readable
```suggestion
    for (let element = 0; element < this.length; element++) {
      if (this[element] !== null || this[element] !== undefined) {
        joinedArray += this[i];
      }

      if (element !== this.length - 1) {
        joinedArray += separator ;
      }
    }

    return joinedArray;
```

ArturErmolenko (changes requested):

ArturErmolenko (approved):

PR #1936, opened 2022-12-30
v-01

AndreyBerezyanskiy:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,29 @@
...
+
+    if (separator === undefined) {
+      stringSeparator = ',';
+    } else {
+      stringSeparator = '' + separator;
+    }
+
+    for (let i = 0; i < copyArray.length; i++) {
+      if (copyArray[i] === null || copyArray[i] === undefined) {
+        copyArray[i] = '';
```
tania-boiar, line 20: 2. [CODE STYLE] - avoid creating needless copies of arrays, you can iterate existing array.

I suppose you don't need here a copy of the array.
You could just create a variable inside a loop for saving an element of the array.

```
      let el = this[i]; 
      if (el === null || el === undefined) {
        el =  '';
        ```

**Review by tania-boiar — COMMENTED** [review]
Good job! 👌

**Review by tania-boiar — APPROVED** [review]

File: src/arrayMethodJoin.js (hunk 1 above)
Comment (AndreyBerezyanskiy) [reply src/arrayMethodJoin.js:20]: Thank you so match! It's useful advice)

**Comment by AndreyBerezyanskiy** [comment]
Thanks!🙂

File: src/arrayMethodJoin.js [hunk 2]
```diff
@@ -5,7 +5,28 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let joinString = '';
+    let stringSeparator = '';
+
+    separator === undefined
+      ? stringSeparator = ','
+      : stringSeparator = '' + separator;
```
Anna-Tkachenko, line 13: set default separator while describing function arguments instead of using separate condition

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      let element = this[i];
+
+      if (element === null || element === undefined) {
+        element = '';
+      }
```
Anna-Tkachenko, line 20: let's add element to the result string only if it's not null or undefined instead of setting empty string to the element

Anna-Tkachenko (changes requested):

AndreyBerezyanskiy:
Thanks! I've fixed it

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinString = '';
```
Anna-Tkachenko, line 8:
```suggestion
    let joinedString = '';
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+    const stringSeparator = separator;
```
Anna-Tkachenko, line 9: do you need this variable?

Anna-Tkachenko (approved):

PR #1937, opened 2022-12-30
solution

imariash86:
my solution

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,32 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let correctSeparator = separator;
+    let str = String(this[0]);
+
+    if (separator === null) {
+      correctSeparator = 'null';
+    }
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    if (str === 'null') {
+      str = '';
+    }
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        str += correctSeparator;
+      } else {
+        str += correctSeparator + this[i];
+      }
+    }
+
+    return str;
   };
 }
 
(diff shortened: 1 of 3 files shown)


nzrii-kysel (approved):
👍

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+      correctSeparator = 'null';
+    }
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    if (str === 'null') {
+      str = '';
+    }
```
ArturErmolenko, line 22: I'm sure that you can do it without this` If`s

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let correctSeparator = separator;
```
ArturErmolenko, line 9: you made default value in function args

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+    let str = String(this[0]);
```
ArturErmolenko, line 10:
```suggestion
    let str = '';
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        str += correctSeparator;
+      } else {
+        str += correctSeparator + this[i];
+      }
+    }
+
+    return str;
```
ArturErmolenko, line 32: avoid `if else `construction 
```suggestion
    for (let i = 0; i < this.length; i++) {
      if (this[i] === undefined || this[i] === null) {
        str += this[i];
      } 
      
      if (i < this.length - 1) {
         str += separator;
       }
    }

    return str;
```

ArturErmolenko (changes requested):

src/arrayMethodJoin.js
imariash86: thanks for your review:)

ArturErmolenko (approved):

PR #1940, opened 2022-12-30
Solution

chermak-oleh:


PolinaSukhorukova (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
```
Anna-Tkachenko, line 8: don't use data type names for variable names
will be better to create more understandable variable name

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1941, opened 2022-12-31
Result

svitjojo:


OlhaYevstifieieva (approved):
Clean Job!

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let string = '';
+    let copySeparator = `${separator}`;
```
alexandra-protyanova, line 9:
```suggestion
    let copySeparator = separator;
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
alexandra-protyanova, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
+      copySeparator = ',';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      switch (typeof this[i]) {
+        case 'object':
+          string += copySeparator;
+          break;
+        case 'undefined':
+          string += copySeparator;
```
alexandra-protyanova, line 21: why do you need these two cases if the are the same?
And I am not sure you will have objesct as elements of array

alexandra-protyanova (changes requested):

ArturErmolenko (approved):

PR #1943, opened 2023-01-01
Solution

tania-boiar:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+
+      if (el === null || el === undefined) {
+        el = '';
+      }
+      string += `${String(separator)}${el}`;
+    }
+
+    return string;
   };
 }
```
AndreyBerezyanskiy, line 30: Nice code! I don't know what to add)

AndreyBerezyanskiy (approved):
Well done!👍

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+    }
+
+    if (this.length === 1) {
+      return String(this[0]);
+    }
+
+    let string = this[0] === null ? '' : String(this[0]);
+
+    for (let i = 1; i < this.length; i++) {
+      let el = this[i];
```
Anna-Tkachenko, line 19: create more understandable variable name instead of `el`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+
+      if (el === null || el === undefined) {
+        el = '';
+      }
```
Anna-Tkachenko, line 23: let's add element to the result string only if it's not null or undefined instead of set empty string here

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
+
+    if (this.length === 1) {
+      return String(this[0]);
+    }
+
+    let string = this[0] === null ? '' : String(this[0]);
```
Anna-Tkachenko, line 16: do you need this?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
+
+    if (this.length === 1) {
+      return String(this[0]);
+    }
```
Anna-Tkachenko, line 14: can we process one element inside the cycle (start it from 0 index) instead of create separate check for the case with 1 element?

Anna-Tkachenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
```
Anna-Tkachenko, line 8: will be better to create more understandable variable name instead of data type name

Anna-Tkachenko (approved):

PR #1945, opened 2023-01-01
Solution

andrii-andriushchenko:


sv-dubov (approved):
Looks good!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let finalStr = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let item = this[i];
+
+      if (item === undefined || item === null) {
+        item = '';
+      }
```
Anna-Tkachenko, line 15: will be better to add the item only if element not undefined or null, it will make your code shorter and clearer

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let finalStr = '';
```
Anna-Tkachenko, line 8: don't use name of data types for variable names

Anna-Tkachenko (changes requested):

src/arrayMethodJoin.js
andrii-andriushchenko: thanks, fixed

src/arrayMethodJoin.js
andrii-andriushchenko: ready

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let finalMethod = '';
```
Anna-Tkachenko, line 8: it's not method 🙂

Anna-Tkachenko (approved):

src/arrayMethodJoin.js
andrii-andriushchenko: I changed again :)

PR #1948, opened 2023-01-02
commit message

jaxxxon78:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let res = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        res += '';
+      } else {
+        res += this[i];
+      }
```
OlEzhA0, line 15:
```suggestion
      if (this[i] !== null || this[i] !== undefined) {
        res += this[i];
      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+
+      if (i !== this.length - 1) {
```
OlEzhA0, line 17:
```suggestion
      if (i < this.length - 1) {
```

OlEzhA0 (changes requested):
Hey! Please check my little suggestions

src/arrayMethodJoin.js
Viktor-Kost: is not fixed. 
No needs to use `else`

Viktor-Kost (changes requested):
Pay attention to unresolved conversation, please

mykhalenych (approved):

PR #1950, opened 2023-01-02
Solution

antskpps:


PolinaSukhorukova (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
```
Anna-Tkachenko, line 8: will be better to create understandable variable name instead of using data type name

Anna-Tkachenko (approved):

PR #1951, opened 2023-01-02
solution_1

Borys-Andrew:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,29 @@
...
     // write code here
+    let result = '';
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    if (this[this.length - 1] === undefined) {
+      this[this.length - 1] = '';
+    }
```
nzrii-kysel, line 17: `join` should not mutate original array

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,29 @@
...
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
+
+      if (separator !== undefined) {
+        result += this[i] + `${separator}`;
+      } else {
+        result += this[i] + ',';
+      }
```
nzrii-kysel, line 28: you do `result += this[i]` in both cases, better take it out of the `if` statement

0xNazarii (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+    }
+
+    if (copy[copy.length - 1] === undefined) {
+      copy[copy.length - 1] = '';
+    }
+
+    for (let i = 0; i < copy.length - 1; i++) {
+      if (copy[i] === null || copy[i] === undefined) {
+        copy[i] = '';
+      }
```
Anna-Tkachenko, line 23: let's add the item to the result string only if it's not empty or undefined instead of setting empty string to the element

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+    const copy = [...this];
+    let result = '';
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    if (copy[copy.length - 1] === undefined) {
+      copy[copy.length - 1] = '';
+    }
```
Anna-Tkachenko, line 18: it's not clear why did you set this check
for what reason did you check the last element of the array?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    const copy = [...this];
```
Anna-Tkachenko, line 9: for what reason did you make the copy of current array?
check do you really need it

Anna-Tkachenko (changes requested):

PR #1954, opened 2023-01-02
Solution

MFrolova-art:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,22 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    const mySeparator = separator === undefined ? ',' : separator;
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === '' || this[i] === undefined) {
```
sv-dubov, line 12: I think you don't need this check: `this[i] === ''` :)

sv-dubov (approved):
Well done!

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,22 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    const mySeparator = separator === undefined ? ',' : separator;
```
Anna-Tkachenko, line 9: let's assign default separator while describing function arguments instead of creating separate check

src/arrayMethodJoin.js
Anna-Tkachenko: agree, fix @sv-dubov comment

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,22 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === '' || this[i] === undefined) {
+        result += '';
```
Anna-Tkachenko, line 13: let's not add empty string here
just add the element if it's not null or undefined

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!

PR #1955, opened 2023-01-02
Solution

VadymTsyndra:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        string += '';
```
agniya-i, line 12: it's not necessary to have this block, leave `string += this[i]` and the condition where you check that `this[i]` is **not** undefined and null

agniya-i (changes requested):

SanyaBratashchuk (approved):

PR #1962, opened 2023-01-03
implemented join method

Bogdan-Kutasevich:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const separatorToString = String(separator);
```
Avichez, line 9: This part is redundant, when you add a separator to string, it will also transform to string.

Avichez (approved):
Good solution 👍
Just a small note.

src/arrayMethodJoin.js
Bogdan-Kutasevich: Thanks for review! I think so, but if i don't used String(separator), it not working when separator is null

ArturErmolenko: if you go from the opposite your code will be more readable

ArturErmolenko, line 9:
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+      if (this[i] === undefined || this[i] === null) {
+        result += separatorToString;
+      } else {
+        result += this[i] + separatorToString;
+      }
+    }
+
+    if (this[this.length - 1]) {
+      result += this[this.length - 1];
+    }
```
ArturErmolenko, line 21:
```suggestion
      if (this[i] !== undefined || this[i] !== null) {
        result += this[i];
      }

    if (i !== this.length - 1) {
         result += separator;
      }
```

ArturErmolenko (changes requested):

PR #1963, opened 2023-01-03
finished join method

andrii-kurash:
Have a good day!✨


andrii-kurash:
> It's very good solution!!!

@Anton-Zhytonbaiev Thanks!)

ArturErmolenko (approved):

PR #1965, opened 2023-01-03
comleted

AlexPro100pro:


PolinaSukhorukova:
Good job!

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,41 @@
...
+      return '';
+    }
+
+    if (separatorDefault === null) {
+      separatorDefault = 'null';
+    }
+
+    if (separatorDefault === undefined) {
+      separatorDefault = ',';
+    }
```
PolinaSukhorukova, line 28: We don't have a condition to check separator by "Null", only by "Undefined", so you can use value by default.

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,41 @@
...
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
     // write code here
+
+    let newStr = `${this[0]}`;
+
+    if (this[0] === undefined || this[0] === null) {
+      newStr = '';
+    }
```
PolinaSukhorukova, line 14: It seems, that you complicate the task when you assign a value to a string as this[0]. If it is empty - you have to do an extra check, in your case - you have 2 extra checks.

PolinaSukhorukova (changes requested):

src/arrayMethodJoin.js
AlexPro100pro: thank, very good idea)

src/arrayMethodJoin.js
AlexPro100pro: I change the logic, it should be fine now.

PolinaSukhorukova (approved):

ArturErmolenko (approved):

PR #1969, opened 2023-01-03
Develop

ann-yurkevych:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
```
nzrii-kysel, line 8: variable name should be at least somewhat meaningful

nzrii-kysel (approved):
👍

PR #1976, opened 2023-01-05
MethodJoin was made

SamirMamedov1010:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,37 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (separator === null) {
```
DarkMistyRoom, line 11: the separator can't be null. you already set default value for it

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,37 @@
...
+        if (i === this.length - 1) {
+          result += this[i];
+          break;
+        }
+
+        result += this[i] + 'null';
+        continue;
+      }
```
DarkMistyRoom, line 19: you can just remove whole this block

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,37 @@
...
+
+      if (i !== this.length - 1) {
```
DarkMistyRoom, line 21: as well as this `if`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,37 @@
...
+          continue;
+        }
+
+        result += this[i] + separator;
+      } else if (i === this.length - 1
+        && this[i] === undefined) {
+        break;
+      } else {
+        result += this[i];
+      }
```
DarkMistyRoom, line 34: refactor this block to make it with only two plane `if`s

ViktorHavryliuk (approved):

PR #1981, opened 2023-01-05
Solution

AV1788:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,17 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let str;
+
+    if (separator !== undefined) {
```
vlad-demchuk, line 10: you could use a default value for functions arguments

vlad-demchuk (changes requested):

loralevitska (approved):
🔥

PR #1993, opened 2023-01-13
solution

Ch0p3r84:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
polosanya, line 8: It is better not to name variables like that. Try to specify what exactly this variable is to make your code cleaner

polosanya (approved):
Well done!

PR #1997, opened 2023-01-16
arr join solution

Spec0s:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+
+      if (string === undefined || string === null) {
+        string = '';
+      }
+
+      if (i === (this.length - 1)) {
+        sourceString += `${string}`;
+      } else {
+        sourceString += `${string}${separator}`;
+      }
```
polosanya, line 22:
```suggestion
      const string = this[i];

      if (string !== undefined && string !== null) {
        sourceString += string;
      }

      if (i !== (this.length - 1)) {
        sourceString += separator;
      }
```
I think you can optimize your solution a bit.
Also, `string` is not the best name, because it's actually just a char

polosanya (changes requested):
Let's polish it up

polosanya (approved):
Great!

Spec0s:
not fully approved ?

polosanya (approved):
What do you mean? This task is already approved)
Write in Slack chat if you have any questions. It is easier to communicate there 😃

PR #2000, opened 2023-01-17
Solution

bojkovladislav:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const copy = [...this];
+    let res = '';
+    let sep = separator + '';
+
+    if (typeof separator === 'number') {
+      return;
+    }
```
ViktorHavryliuk, line 14: No need to check type if you already converted it
```suggestion
    const separatorString = String(separator);
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const copy = [...this];
```
ViktorHavryliuk, line 8: Can be done without copy
```suggestion
```

ViktorHavryliuk (changes requested):

tetiana-ivanchenko (approved):

PR #2001, opened 2023-01-18
Solution

kirill-vypirovskyi:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1) {
+        separator = '';
+      }
+
+      if (this[i] === null || this[i] === undefined) {
+        result = result + separator;
+      } else {
+        result = result + this[i] + separator;
+      }
```
kuznetsova-anastasiia, line 20: checklist pt.3
```suggestion
      if (this[i] === null || this[i] === undefined) {
        result += separator;
      } else {
        result += this[i] + separator;
      }
```

kuznetsova-anastasiia (changes requested):
GJ, but i found one inconsistency with the checklist

Andr1yK (approved):

PR #2006, opened 2023-01-20
Solution

OlyaNikora:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+    const last = this[this.length - 1];
+
+    for (const item of this) {
+      if (item === null || item === undefined) {
+        string += '';
```
mykhalenych, line 13: I think it is redundant

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+      } else {
+        string += item;
+      }
```
mykhalenych, line 16:
```suggestion
      if (item !== null || item !== undefined) {
        string += item;
```

mykhalenych (changes requested):

src/arrayMethodJoin.js
OlyaNikora: I tried your option, but 4 tests are failed. Back to my solution.

src/arrayMethodJoin.js
OlEzhA0, line 16: just use
```suggestion
      if (item !== null && item !== undefined) {
        string += item;
      }
```

OlEzhA0 (changes requested):
hey! just added small improvements

src/arrayMethodJoin.js
OlyaNikora: I tried once more => 4 tests are failed. 
image

src/arrayMethodJoin.js
witflash: Because correct check will be:

```
        string += item;
      }
```

src/arrayMethodJoin.js
witflash: Look carefully - not `||` but `&&`.
And all tests will pass.

witflash (changes requested):
Carefully check the last comments, please.

src/arrayMethodJoin.js
OlyaNikora: Thank you! :)  Done

IvanFesenko (approved):

PR #2007, opened 2023-01-20
solution

44shaggy:


src/arrayMethodJoin.js
```diff
@@ -3,9 +3,24 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] == null) {
+        string += ``;
+      } else {
```
polosanya, line 14:
```suggestion
      if (this[i] !== null) {
```
Can't you just do nothing?
Also, always use === instead of ==
image

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comments

src/arrayMethodJoin.js
44shaggy: I have a question though. I used
(value != null) 
as a shorthand for
(value !== null && value !== undefined)
Is it wrong?

PR #2017, opened 2023-01-25
Solution

nadia-hope:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (const arg of this) {
+      if (arg === null || arg === undefined) {
+        string += '';
+      } else {
+        string += arg;
+      }
```
agniya-i, line 15: let's improve this part a little bit - you can leave only `if` block and check if the argument is not null **and** not undefined  

```suggestion
      if (arg !== null && arg !== undefined) {
        string += arg;
       }
```

agniya-i (changes requested):

polosanya (approved):
Good job!

PR #2019, opened 2023-01-26
function was created

YevheniiHura:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+        this[i] = '';
+      }
+      result += this[i];
+      result += separator;
+    }
+
+    if (typeof this[this.length - 1] === 'undefined'
+    || this[this.length - 1] === null) {
+      this[this.length - 1] = '';
+    }
```
IvanFesenko, line 25: Why you need this?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+    if (this.length === 0) {
+      return result;
+    }
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (typeof this[i] === 'undefined' || this[i] === null) {
+        this[i] = '';
+      }
+      result += this[i];
+      result += separator;
```
IvanFesenko, line 19: You are mutate original array. Do action if `this[i] !== null && this[i] !== undefined
`
```suggestion
      result += `${this[i]}${separator}`;
```

IvanFesenko (changes requested):
Good job. Some fix required

src/arrayMethodJoin.js
YevheniiHura: done

src/arrayMethodJoin.js
YevheniiHura: I've made some corrections, but i need this to check the last element of the array, it might be null or undefined. Can I check it in another way?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+    if (this.length === 0) {
+      return result;
+    }
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        result += `${this[i]}${separator}`;
+      } else {
+        result += separator;
+      }
```
Bohdan-Maksymiuk, line 19: no need else here - add separator in every case of this is not the last element

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+    }
+
+    if (this[this.length - 1] !== undefined && this[this.length - 1] !== null) {
+      result += this[this.length - 1];
+    }
```
Bohdan-Maksymiuk, line 24:
```suggestion
      if (i < this.length - 1) {
        result += separator;
      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (this.length === 0) {
+      return result;
+    }
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        result += `${this[i]}${separator}`;
```
Bohdan-Maksymiuk, line 16:
```suggestion
        result += `${this[i]}`;
```

Bohdan-Maksymiuk (changes requested):

src/arrayMethodJoin.js
YevheniiHura: done

src/arrayMethodJoin.js
YevheniiHura: I'm sorry, can you please explain me how to apply this part of the code? I don't have variable 'i' (except for index in the loop before) and I don't understand how it can help me to check the last element of the array.

IvanFesenko: You need do this check inside loop. If you have more questions, ask them in fe_chart

IvanFesenko (changes requested):
Not fixed

src/arrayMethodJoin.js
YevheniiHura: Thank you for the hint, done)

PR #2020, opened 2023-01-26
commit message

Peti51:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
jstmpelowycz, line 7: better to set default separator `,`

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this.length === 1) {
+        joined += this[i];
+        break;
+      }
+
+      if (isNaN(separator)) {
+        joined += this[i] + ',';
+      }
```
jstmpelowycz, line 18: actually, we don't need it

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let joined = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this.length === 1) {
+        joined += this[i];
+        break;
+      }
```
jstmpelowycz, line 14: this one is redundant, no need to do it (there's gonna be just 1 iteration in such cases, we won't go forward)

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
+      } else if (i + 1 === this.length) {
+        joined += this[i];
+        break;
+      }
+
+      if (this[i] === null || this[i] === undefined) {
+        joined += '' + separator;
+      } else if (this[i] !== null || this[i] !== undefined) {
+        joined += this[i] + separator;
+      }
```
jstmpelowycz, line 31: too complicated

jstmpelowycz (changes requested):
what you actually need – on each iteration, you have to make sure the current element is not `null` or `undefined` and concatenate it to `joined`, separator can be added separately, till the pre-last element (as far as we don't need separator at the end)
```
if neither null nor undefined –> concatenate

if not the pre-last item –> add separator
```
try it out!

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,26 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
polosanya, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```
You can just do like this

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,26 @@
...
+        joined += this[i];
+      }
+
+      if (i !== this.length - 1 && separator !== undefined) {
+        joined += separator;
+      }
+
+      if (i !== this.length - 1 && separator === undefined) {
+        joined += coma;
+      }
```
polosanya, line 22:
```suggestion
    for (let i = 0; i < this.length; i++) {
      if (this[i] !== null && this[i] !== undefined) {
        joined += this[i];
      }

      if (i !== this.length - 1) {
        joined += separator;
      }
```

polosanya (changes requested):
Let's polish it up)

qoa11a (approved):

PR #2022, opened 2023-01-27
task was solved

vanyachyzh:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if ((this[i] === undefined && i === this.length - 1)
+      || (this[i] === null && i === this.length - 1)) {
```
witflash, line 12: Why do you need these checks `i === this.length - 1`? 
I think you don't need whole this block.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+        result += '';
+      } else if (this[i] === undefined || this[i] === null) {
+        result += '' + (separator + '');
```
witflash, line 15: Why do you need to add an empty string at the end? There is no sense to do this.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+      } else if (i === this.length - 1) {
+        result += this[i];
+      } else {
+        result += this[i] + (separator + '');
```
witflash, line 19: And here you don't need to add an empty string too.

witflash (changes requested):
You can do it much easier.
Just check `if (this[i] !== undefined && this[i] !== null)` then add `this[i]` to the `result`.
And then you can check if it's not the last element and add `separator` to the end.

So you will have just two `if` blocks without `else`.

src/arrayMethodJoin.js
vanyachyzh: My function doesn't work in right way when null is passed as argument. 
[0, 1, 2, 3].join(null)
Expected: "0null1null2null3"
Received: "0123"
So I decided to solve this problem with (separator + '')

vanyachyzh: If I don't know how to solve this issue with null I need to go to chat and ask there or I can ask question about that here?)) How to ask correctly?

src/arrayMethodJoin.js
vanyachyzh: (i === this.length - 1) I check if it element is the last. If it is true I don't add separator

src/arrayMethodJoin.js
vanyachyzh: I got it!  
I need  (separator + '') in order to turn null into string
0 + null = 0  => number + number = number => "0123"
0 + (null + '') = 0null => number + (number + string) = number + string = string => "0null1null2null3"

VitaliyBondarenko1982 (approved):

VitaliyBondarenko1982 (approved):

PR #2023, opened 2023-01-28
the task was solved

batiukova-nataliia:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    let sep = String(separator);
+
+    if (separator === undefined) {
+      sep = ',';
+    }
```
tetiana-ivanchenko, line 13: use default parametr instead of this block

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        this[i] = '';
+      }
+
+      if (i === this.length - 1) {
+        sep = '';
+      }
+      result += `${this[i] + sep}`;
```
tetiana-ivanchenko, line 23:
```suggestion
      if (this[i] !== undefined || this[i] !== null) {
        result += this[i];
      }

      if (i !== this.length - 1) {
        result += sep
      }
```

tetiana-ivanchenko (changes requested):

src/arrayMethodJoin.js
batiukova-nataliia: Tests don't pass with this code. I changed it another way with mentors' suggestions.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const sep = String(separator);
```
etojeDenys, line 9: you can just use `separator` instead of `sep`

etojeDenys (approved):
LGTM! 👍🏻
A few comments.

VitaliyBondarenko1982 (approved):
Good job!

PR #2024, opened 2023-01-28
joinMethod

RomanHrytsiuk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+
+    if (this.length === 0) {
+      return '';
+    }
```
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+        str += separator;
+        continue;
+      }
+      str += this[i];
+      str += separator;
+    }
+
+    if (this[this.length - 1] !== undefined) {
+      str += this[this.length - 1];
+    }
```
SanyaBratashchuk, line 25:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        str += this[i];
      }
      
    if (i !== this.length - 1) {
      str += separator;
    }
```

SanyaBratashchuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+      }
+
+      if (i !== this.length - 1) {
+        str += separator;
+      }
+    }
+
+    if (this[this.length - 1] !== undefined) {
+      str += this[this.length - 1];
+    }
```
DarkMistyRoom, line 22: why do you need this condition? It looks redundant

DarkMistyRoom (changes requested):

src/arrayMethodJoin.js
RomanHrytsiuk: The code doesn't work without the lines 20-22

DarkMistyRoom: for the future: if you don't agree or have questions, write in the chat. it will be faster

DarkMistyRoom, line 22:
```suggestion
    for (let i = 0; i < this.length; i++) {
      if (this[i] !== null && this[i] !== undefined) {
        str += this[i];
      }

      if (i !== this.length - 1) {
        str += separator;
      }
    }
```

DarkMistyRoom: now it should work

DarkMistyRoom (approved):

PR #2025, opened 2023-01-29
Solution

G37ua:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let arr = '';
```
polosanya, line 8: It's not array. Let's use more descriptive names

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,24 @@
...
+    let separatorN = separator;
+
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1) {
+        separatorN = '';
+      }
```
polosanya, line 14: Just add a separator only if it's not a last element

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,24 @@
...
+
+      if (this[i] !== undefined && this[i] !== null) {
+        arr += this[i];
```
polosanya, line 17: You can create a variable for `this[i]` to make your code a bit cleaner

polosanya (changes requested):

G37ua:
Исправил

Anastasiia-Svintsova (approved):

PR #2033, opened 2023-02-01
Solution

katiailchenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,42 @@
...
+        break;
+      };
+
+      if (item === undefined || item === null) {
+        string += String(separator);
+
+        continue;
+      };
+
+      string += item + String(separator);
```
ivadimko, line 39: seems like String(separator) is repeated several times

ivadimko (approved):

PR #2044, opened 2023-02-05
Method Join solution

RomanTsvindin:
version 1

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,9 +4,37 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let separatorString;
+    let finalString = '';
+
+    switch (separator) {
+      case null: separatorString = 'null';
+        break;
+      case undefined: separatorString = 'undefined';
+        break;
+      case '': separatorString = '';
+        break;
+      default: separatorString = separator;
+    }
+
+    for (const element of this) {
+      switch (element) {
+        case undefined: finalString += '';
+          break;
+        case null: finalString += '';
+          break;
+        case NaN: finalString += 'NaN';
+          break;
+        default: finalString += element;
+      }
+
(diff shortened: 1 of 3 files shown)


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,37 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let separatorString;
+    let finalString = '';
+
+    switch (separator) {
+      case null: separatorString = 'null';
+        break;
+      case undefined: separatorString = 'undefined';
+        break;
+      case '': separatorString = '';
```
polosanya, line 16: You can just check if it's not null and undefined, and add a separator to the string only in that case.
There is really no need to use switch case

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,37 @@
...
+
+    for (const element of this) {
+      switch (element) {
+        case undefined: finalString += '';
+          break;
+        case null: finalString += '';
+          break;
+        case NaN: finalString += 'NaN';
+          break;
+        default: finalString += element;
```
polosanya, line 29: Same here. You don't need to add empty string and NaN manually. Those values will be added anyway

polosanya (changes requested):
Good job! Let's try to improve your code a bit)

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,25 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let separatorString;
+    let finalString = '';
+
+    if (separator !== ('null' || 'undefined')) {
+      separatorString = separator;
+    }
```
polosanya, line 13:
```suggestion
    let finalString = '';
```
Why do you need this logic? Just use `separator` instead of `separatorString`

polosanya (changes requested):
Almost perfect!

anstsot (approved):
Great job, approved!

PR #2047, opened 2023-02-07
Array method Join solution

TkachYuliia:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      let element = this[i];
+
+      if (element === null || element === undefined) {
+        element = '';
+      }
+
+      result += element;
```
IvanFesenko, line 21: This one looks better. We do action when we need, no need to overwrite to empty string
```suggestion
      const element = this[i];

      if (element !== null || element !== undefined) {
        result += element;
      }
```

IvanFesenko (changes requested):
Good job. Check my comment

src/arrayMethodJoin.js
TkachYuliia: At least 4 tests are not working if I rewrite the code this way. So I changed it in a bit different way. Here is one of the unsuccessful tests - https://prnt.sc/ZqrOJ81I-c1S

etojeDenys: hmm, that sounds strange, can you please share a photo of your code after these changes?

TkachYuliia: I'm so sorry, it seems that I missed something on the first try because on the second try everything went well. So it's my mistake, sorry.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (this.length === 0) {
+      return result;
+    }
```
IvanFesenko, line 12: BTW this check is redudant.

IvanFesenko (approved):

PR #2049, opened 2023-02-07
task was solved

anton-chornobai:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (this[this.length - 1] === undefined) {
+      this[this.length - 1] = '';
+    }
```
IvanFesenko, line 12: Return empty string if array is empty.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
```
IvanFesenko, line 17:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        result += this[i];
      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+      result += this[i] + `${separator}`;
+    }
```
IvanFesenko, line 19: You should add separator except last item

IvanFesenko (changes requested):

src/arrayMethodJoin.js
anton-chornobai: But if element is null or undefined it should be replaced with ' ' . 

Expected: ",,1,,0,NaN,"
    Received: "1,0,NaN,,,"

      75 | test(`Should work when array has null as a first element`, () => {
      76 |   expect([null, null, 1, undefined, 0, NaN, ''].join2(','))
    > 77 |     .toBe(',,1,,0,NaN,');
         |      ^

etojeDenys (approved):
LGTM

PR #2050, opened 2023-02-07
Solution

8Melya8:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
...
+    let str = '';
+
+    if (separ === undefined) {
+      separ = ',';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
```
IvanFesenko, line 18: We can do action in case when has value, no need to rewrite
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        str+= this[i];
      }
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let separ = separator;
+    let str = '';
```
IvanFesenko, line 9: remove `separ` variable, add default value to args
`function(separator = ',')`

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
+
+      if (i === this.length - 1) {
+        str += this[i];
+      } else {
+        str += `${this[i]}${separ}`;
+      }
```
IvanFesenko, line 24: add separator only when `i < this.length - 1`. 
```suggestion
      if (i < this.length - 1) {
        str += separator
      } 
```

IvanFesenko (changes requested):
Good job. Some refactor required

tetiana-ivanchenko (approved):

PR #2051, opened 2023-02-08
solution

ivanm2706:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+        } else {
+          string += `${this[i]}`;
+        };
+      } else {
+        if (i !== this.length - 1) {
+          string += separator;
+        } else {
+          string += '';
+        };
+      }
```
IvanFesenko, line 23: This is redudant

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        if (i !== this.length - 1) {
+          string += `${this[i]}${separator}`;
+        } else {
+          string += `${this[i]}`;
+        };
```
IvanFesenko, line 16: make to if statement:
first when we add next item
second - we add separator until it's not last item

IvanFesenko (changes requested):
Good job. Some refactor needs

mykhalenych (approved):

PR #2053, opened 2023-02-08
Solution

Vasyl-Pauk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joined = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
```
OleziO, line 11: `if(this[0])` its same as `if(this[i] !== null && this[i] !== undefined)`

OleziO (approved):

PR #2054, opened 2023-02-08
Solution

valeriaoleynik:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const sep = separator;
```
witflash, line 9: Why do you need this additional constant? You could use `separator` below directly.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+
+    if (this.length === 0) {
+      return result;
+    }
```
witflash, line 13: There is no need to check it. If you will have an empty array your cycle will not do anything and the function will return the result immediately.

witflash (changes requested):
Good work.
But check my comments, please. And do small fixes, please.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
polosanya, line 8: Use more descriptive names than just "result"

polosanya (approved):
Good job!

PR #2062, opened 2023-02-10
Task done

sergik94:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    if (typeof separator === 'number') {
```
mykhalenych, line 10: you already added default param it can't be number

mykhalenych (changes requested):

src/arrayMethodJoin.js
sergik94: I misunderstood one condition

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+      return string;
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        string += this[i];
+      }
+
+      if (i !== this.length - 1) {
+        string += `${separator}`;
```
etojeDenys, line 20:
```suggestion
        string += separator;
```

etojeDenys (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    if (this.length === 0) {
```
VitaliyBondarenko1982, line 10:
```suggestion
    if (!this.length) {
```

VitaliyBondarenko1982 (approved):
Good job.

Just one liitle suggestion about checking this.length,
but anyway approve.

PR #2064, opened 2023-02-11
solution

MSA1991:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const arr = this;
```
Volodymyr-Melnyk-Fron10der, line 8: It's actually useless variable, you can access array with `this ` right away

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+    let result = '';
+
+    for (let i = 0; i < arr.length; i++) {
+      if (arr[i] !== null && arr[i] !== undefined) {
```
Volodymyr-Melnyk-Fron10der, line 12: As I know eslint won't allow you to do it, but just FYI:

There is operator such `??` which would be great to use here

Volodymyr-Melnyk-Fron10der (changes requested):
Great👍 Just remove arr variable and work with `this` to not waste memory

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      // eslint won't let me do it
+      // this[i] = this[i] ?? '';
+      // result += this[i];
```
Volodymyr-Melnyk-Fron10der, line 13: Got it, as I said before) No problem

Volodymyr-Melnyk-Fron10der (approved):
Good job✅
But it`s better to not commit any comments in code, write them as a PR comments

PR #2067, opened 2023-02-12
join solution

kateryna-babina:


vladyslav-mosor (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,34 @@
...
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    let fixedSeparator = separator;
+
+    if (separator === null) {
+      fixedSeparator = 'null';
+    }
```
illiarerun, line 18: why do you need it? null will easily transform into string
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,34 @@
...
+
+    for (let i = 0; i < this.length - 1; i++) {
+      let element = this[i];
+
+      if (element === undefined || element === null) {
+        element = '';
+      }
+
+      strJoin += element + fixedSeparator;
```
illiarerun, line 27: let's improve it a bit, instead of introducing `let` variable, you can invert your condition and just add element beforehand, like:
```suggestion
      if (condition where you SHOULD add element) {
        // here you add element to your string
      }

      // here you add separator to it
```

illiarerun (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let strJoin = '';
+
+    if (this.length === 0) {
+      return '';
+    }
```
OleksiiBlezniuk, line 12: This condition is unnecessary, your loop will not start if array is empty

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+      if (element !== undefined && element !== null) {
+        strJoin += element;
+      }
+
+      strJoin += separator;
+    }
+
+    if (this[this.length - 1]) {
+      strJoin += this[this.length - 1];
+    }
```
OleksiiBlezniuk, line 26: Move it to loop and iterate through all elements
Just don't add separator if it's a last element of array

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let strJoin = '';
```
OleksiiBlezniuk, line 8: Bad variable name, it should contain result of your logic, so `joined` will be more clear
Also better to avoid using data type in variable name

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
Well done!

PR #2070, opened 2023-02-13
solution Method join

Zelinskyi-Serhii:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    let between = separator;
+
+    if (separator === undefined) {
+      between = ',';
+    }
```
rashyd-hasratov, line 13: Actually, you don't need this part of code, you can just set the default value for separator in the function params list.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
...
+        continue;
+      }
+
+      if (i + 1 !== this.length) {
+        result += `${this[i]}${between}`;
+        continue;
+      }
+
+      if (this[this.length - 1] !== undefined) {
+        result += this[i];
```
rashyd-hasratov, line 27: I would like to offer you such simplification for this part of code:
1) if it is not the first element, add separator to result
2) if it is null or undefined, just skip this iteration
3) else add the element to the result

rashyd-hasratov (changes requested):
Nice, but it is possible to make the solution even better!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const between = separator;
```
OleksiiBlezniuk, line 9: But why do you need this constant? You already have `separator`

OleksiiBlezniuk (changes requested):
Well done, but let's avoid using unnecessary constants

OleksiiBlezniuk (approved):
Well done!

PR #2076, opened 2023-02-15
solved

danulyk05:


ViktorHavryliuk (changes requested):
There are too many excessive actions and checks, this code can be much simpler and shorter. Try to figure it out on your own as much as possible, but if you need help feel free to ask in chat

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,34 @@
...
+
+    if (this.length === 0) {
+      return '';
+    };
+
+    if (separators === null) {
+      separators = 'null';
+    };
+
+    if (this[0] === null) {
```
OleziO, line 19: Redundant code, this check is not needed.

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,34 @@
...
+      this[0] = '';
+    }
+
+    str += this[0];
```
OleziO, line 23: Redundant code, this check is not needed.

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,34 @@
...
+
+    for (let i = 1; i < this.length; i++) {
```
OleziO, line 25: Start from 0 index.

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,34 @@
...
+      if (this[i] === undefined || this[i] === null) {
```
OleziO, line 26: Replace with `!this[i]`, its the same as `this[i] === undefined || this[i] === null`

OleziO (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,25 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (!this.length) {
+      return '';
+    };
```
nazarmatsevych, line 12: redundant check, you already have this check in your loop

nazarmatsevych (changes requested):

mykhalenych (approved):

PR #2080, opened 2023-02-15
join solution

Yekatery:


NikitaStambul (approved):
good job

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
```
illiarerun, line 8: dnt tr t mk r cd shrtr b rmvng vwls. But if I wanted you to understand what I wrote, I would write: "don't try to make your code shorter by removing vowels "

illiarerun (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
```
OleksiiBlezniuk, line 8:
```suggestion
    let result = '';
```

OleksiiBlezniuk: This will be enough, better to avoid using data type in variable name

OleksiiBlezniuk (approved):

PR #2082, opened 2023-02-15
Solution

andrii-kovalskyi02:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
```
OleziO, line 5: Delete comment.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
```
OleziO, line 11: Replace with `this[i]`, its the same as` this[i] !== null && this[i] !== undefined`

OleziO (approved):

src/arrayMethodJoin.js
andrii-kovalskyi02: It should be like that because not all falthy values should be ignored.

PR #2087, opened 2023-02-16
solution

NikitaStambul:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const strSeparator = String(separator);
+    let result = '';
+
+    for (const el of this) {
+      let strEl = el;
+
+      if (el === undefined || el === null) {
```
gretoq, line 14: Why you use el, if you created variable "strEl"

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const strSeparator = String(separator);
+    let result = '';
+
+    for (const el of this) {
+      let strEl = el;
```
gretoq, line 12: Strange variable name "strEl"

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const strSeparator = String(separator);
+    let result = '';
+
+    for (const el of this) {
```
gretoq, line 11: "el" is also strange name for item of "this"

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+      let strEl = el;
+
+      if (el === undefined || el === null) {
+        strEl = '';
+      }
+
+      result += strEl + strSeparator;
+    }
+
+    return result.slice(0, result.length - strSeparator.length);
```
gretoq, line 21: I don't sure can we use "slice" method in this task, but we can solve this task without "slice" method.

gretoq (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const strSeparator = String(separator);
```
OleksiiBlezniuk, line 8: Why do you need to create this constant? You'll always add this separator to string and it will be converted to sting by default

'' + null === 'null'
'' + undefined === 'undefined'

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
...
+    let result = '';
+
+    for (const str of this) {
```
OleksiiBlezniuk, line 11: Why `str`? What if it will be array of numbers?

OleksiiBlezniuk: Use more common name here, e.g. `element`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
...
+      if (str === undefined || str === null) {
+        result += strSeparator;
+      } else {
+        result += str + strSeparator;
+      }
```
OleksiiBlezniuk, line 16: In both cases you'll add separator, let's move this action outside of if/else block

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
...
+    }
+
+    return result.slice(0, result.length - strSeparator.length);
```
OleksiiBlezniuk, line 19: Let's not add redundant separator instead of adding and then removing it

OleksiiBlezniuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const el = this[i];
```
sergii-nosachenko, line 11: What does `el` mean? `elephant`, `electron`? Shortening vars is not an optimization.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+
+      if (!(el === undefined) && !(el === null)) {
+        result += el;
+      }
```
sergii-nosachenko, line 15: Could be simplified using [Nullish coalescing operator (??)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing).

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+
+      if (!(i === this.length - 1)) {
```
sergii-nosachenko, line 17: Why do you need NOT operator? We have `!==` in JS for such a condition. Or even better - `<`.

sergii-nosachenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const element = this[i];
+
+      if (!(element === undefined) && !(element === null)) {
```
sergii-nosachenko, line 13: Use `!==` instead of the NOT operator.

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #2090, opened 2023-02-16
Solution

nazarkharkevych:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let element = this[i];
+
+      if (this[i] === null || this[i] === undefined) {
```
OleziO, line 13: Replace with `!this[i]`, Its the same as `this[i] === null || this[i] === undefined`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+        element = '';
+      }
+
+      if (i !== this.length - 1) {
```
OleziO, line 17: Better change `i !== this.length - 1` to
`i < this.length - 1`

OleziO (approved):

src/arrayMethodJoin.js
nazarkharkevych: It's not the same in this case, if replaced, the function fails the tests because the loop ignores 0, and NaN.

ViktorHavryliuk (approved):

PR #2091, opened 2023-02-16
join method recreated

andriibezkorovainyi:


NataliaSavenets (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedArr = '';
+    let pseudoSeparator = separator;
+
+    if (separator === null) {
+      pseudoSeparator = 'null';
+    }
+
+    for (const n of this) {
+      let item = n + '';
```
illiarerun, line 16: why do you need it?
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+
+      if (typeof n === 'undefined' || n === null) {
```
illiarerun, line 18: it'll never be equal to `null` after the conversion to string, so that's why you should remove this `let item = n +''` code

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+        item = '';
+      }
+
+      if (this.lastIndexOf(n) === this.length - 1) {
+        joinedArr += item;
+        break;
+      }
+      joinedArr += item + pseudoSeparator;
```
illiarerun, line 26: looks too complicated, can you add item before your check and the decide in if if you should add the separator? cause you're adding item anyways. then you can remove your `break`)

illiarerun (changes requested):

src/arrayMethodJoin.js
andriibezkorovainyi: But there is a test, saying that it shouldn't add new items if they are null (not 'null'). How to check on null without `n === null`?
image

andriibezkorovainyi: Just filter the array after?🤔

andriibezkorovainyi:
@illiarerun I didn't figure out how to omit `n === null` check. I've implemented the rest of your suggestions.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringFromThis = '';
```
OleksiiBlezniuk, line 8: Bad variable name, it should describe what is stored in this variable (result of `join` method)
Also better to avoid using data type in variable name

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+    const pseudoSeparator = separator;
+
+    for (const n of this) {
+      let item = n;
+
+      if (n === undefined || n === null) {
+        item = '';
+      }
+
+      stringFromThis += item;
```
OleksiiBlezniuk, line 18: Or you can just don't add nothing if this element is null or undefined
(adding empty string and doing nothing will have same result)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+
+      if (this.indexOf(n) !== this.length - 1) {
+        stringFromThis += pseudoSeparator;
+      }
```
OleksiiBlezniuk, line 22: Why to search for index of element (and iterate your array to get it) if you can write `for  in` loop and always have index?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultOfJoinMethod = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        resultOfJoinMethod += this[i];
+      }
```
sergii-nosachenko, line 13: A good place to use [the nullish coalescing operator (??)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing)

sergii-nosachenko (approved):

PR #2093, opened 2023-02-16
join method first try

RufatRM:


src/arrayMethodJoin.js
```diff
@@ -1,11 +1,41 @@
 'use strict';
 
+const nullString = 'null';
+const emptyString = '';
```
Volodymyr-Melnyk-Fron10der, line 4: Keep rules of constant naming
```suggestion
const NULL_STRING= 'null';
const EMPTY_STRING= '';
```

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,41 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+
+    if (this.length === 0) {
+      return emptyString;
+    } else if (this.length === 1) {
+      return this[0].toString();
+    }
```
Volodymyr-Melnyk-Fron10der, line 17: use early return pattern

```suggestion
    if (this.length === 0) {
      return emptyString;
    } 
// By the way, do you really need it? Wouldn't the method work without this condition?
    if (this.length === 1) {
      return this[0].toString();
    }
```

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,41 @@
...
+
+    let result = this[0];
+
+    if (result === null) {
+      result = emptyString;
+    }
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = emptyString;
```
Volodymyr-Melnyk-Fron10der, line 33: join is a non-mutating method

Volodymyr-Melnyk-Fron10der (changes requested):
Hi, keep it up👍

src/arrayMethodJoin.js
RufatRM: Hi, I didn't understand what should be changed here,

polosanya (changes requested):
Hi, I see you haven't fixed previous comments, maybe you did not push your changes?
If you have any questions, please ask them in Slack chat and tag mentors. It will be faster because your answers in GitHub will be visible to the mentor only after you re-request review 🙂

Volodymyr-Melnyk-Fron10der (approved):
Looks good👍

PR #2094, opened 2023-02-17
solution

abcdesign1:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',' || 'null') {
+    if (this.length === 0) {
+      return '';
+    }
```
stanislavkulakovskyi, line 10: this condition is unnecessary

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',' || 'null') {
```
stanislavkulakovskyi, line 7: why do you need this **|| 'null'** here?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+      return '';
+    }
+
+    let joined = '';
+
+    let newSeparator = separator;
+
+    if (separator === null) {
+      newSeparator = 'null';
+    }
```
stanislavkulakovskyi, line 18: this one is unnecessary as well

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+      if (i === this.length - 1) {
+        joined += this[i] || '';
+        break;
+      }
+
+      if (this[i] === undefined || this[i] === null) {
+        joined += '' + newSeparator;
+        continue;
+      }
+      joined += this[i] + newSeparator;
```
stanislavkulakovskyi, line 30: I guess it's better to write condition something like this:

        joined += this[i];
      }

      if (i !== this.length - 1) {
        joined += separator;
      }
      
this way it looks more clear and compact

stanislavkulakovskyi (changes requested):
good job overall, but there are some things that could be improved!

src/arrayMethodJoin.js
vadym4che: fixed

src/arrayMethodJoin.js
vadym4che: fixed

src/arrayMethodJoin.js
vadym4che: fixed

stanislavkulakovskyi (approved):
Well done mate!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        joinedString += this[i];
+      }
+
+      if (i !== this.length - 1) {
```
OleksiiBlezniuk, line 15:
```suggestion
      if (i < this.length - 1) {
```

OleksiiBlezniuk: This way is more logical as for me

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
```
OleksiiBlezniuk, line 8: Ok in this context but better to avoid using data type in variables names

OleksiiBlezniuk (approved):

PR #2097, opened 2023-02-17
Created a "join" method for creating a string from an array

gretoq:


arkushenko-anna (approved):
Good Job!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+    let joinedToString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let itemOfThis = this[i];
+
+      if (itemOfThis === null || itemOfThis === undefined) {
+        itemOfThis = '';
+      }
+
+      joinedToString += itemOfThis;
```
illiarerun, line 17: you can invert your check so you'd only do one thing
```suggestion
      if (inverted check here) {
        joinedToString += itemOfThis;
      }
```

illiarerun (approved):

PR #2098, opened 2023-02-17
created program of join

vadosik15:


Lilu580 (commented):
Good solution

Vlados3611 (approved):
Good solution!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let newstr = '';
```
OleksiiBlezniuk, line 8: Bad variable name, it's absolutely not clear what is stored here

OleksiiBlezniuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
```
sergii-nosachenko, line 8: Why `string`? Describe what string it is - `joinedArray`, `joinedString`...

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (i !== 0 && i < this.length) {
+        string += separator;
+      }
```
sergii-nosachenko, line 13: You can add a separator after the element. It needs only to check `i` is less than `length - 1`.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+
+      if (this[i] !== undefined && this[i] !== null) {
+        string += this[i];
+      }
```
sergii-nosachenko, line 17: Could be simplified using [Nullish coalescing operator (??)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing).

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #2103, opened 2023-02-17
Solution

arkushenko-anna:


olecsiy-gavrilyuk (approved):
GJ

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let newString = '';
```
OleksiiBlezniuk, line 8: Bad variable name, it's not clear what `new` mean

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if ((this[i] !== undefined) && (this[i] !== null)) {
```
OleksiiBlezniuk, line 11:
```suggestion
      if (this[i] !== undefined && this[i] !== null) {
```

OleksiiBlezniuk: You don't need this brackets here

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+        newString += this[i];
+      }
+
+      if (i !== this.length - 1) {
```
OleksiiBlezniuk, line 15:
```suggestion
      if (i < this.length - 1) {
```

OleksiiBlezniuk: This way is more logical as for me

OleksiiBlezniuk (changes requested):
GJ, but there are few small issues to fix

sergii-nosachenko (approved):
Good work 👍

PR #2104, opened 2023-02-17
sol

Lilu580:


src/arrayMethodJoin.js
```diff
@@ -1,11 +1,20 @@
 'use strict';
 
-/**
- * Implement method join
- */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let res = '';
```
kuzmi491, line 5: mentors said not to use 'res' =))

kuzmi491 (approved):

src/arrayMethodJoin.js
OleksiiBlezniuk, line 5: Ok, but it's a bad variable name, in this case `result` will be ok, but better to avoid words trimming

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,20 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        res += this[i];
+      }
+
+      if (i !== this.length - 1) {
```
OleksiiBlezniuk, line 12:
```suggestion
      if (i < this.length - 1) {
```

OleksiiBlezniuk: That way is more logical as for me

OleksiiBlezniuk (approved):

PR #2109, opened 2023-02-18
solution

siddhadatri:


Stepan-TS (approved):
Good job! But I think that the solution can be shorter)
1.You can use default value for argument of function, then you don't need check if value === undefined;
2. You double check that the value === null, but you can check in your loop if current !== null and !== undefined, then you don't need create new variable newSeparator;

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,32 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
sergii-nosachenko, line 7: Add default value for function argument

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,32 @@
...
+    let result = '';
+    let newSeparator = separator;
+
+    if (newSeparator === undefined) {
+      newSeparator = ',';
+    }
+
+    if (newSeparator === null) {
+      newSeparator = 'null';
+    }
```
sergii-nosachenko, line 17: Useless code

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,32 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      let current = this[i];
+
+      if (current === null || current === undefined) {
+        current = '';
+      }
```
sergii-nosachenko, line 24: Instead of checking for a value that is `null` or `undefined`, better check the opposite - not equal to these values. Then you can do `result += current;` directly here.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,32 @@
...
+
+      if (i !== this.length - 1) {
```
sergii-nosachenko, line 26:
```suggestion
      if (i < this.length - 1) {
```
More clear way

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,32 @@
...
+        result += current + newSeparator;
+      } else {
+        result += current;
+      }
```
sergii-nosachenko, line 30: 1. Just add a separator
2. Useless `else` branch

sergii-nosachenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    let newSeparator = separator;
```
sergii-nosachenko, line 9: You don't need this variable

sergii-nosachenko (changes requested):

Andr1yK (approved):

PR #2115, opened 2023-02-19
join method

Ambulence:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let newStr = '';
```
sergii-nosachenko, line 8: Use meaningful var name.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        newStr += this[i];
+      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+
+      if (i !== this.length - 1) {
```
```suggestion
      if (i < this.length - 1) {
```

It is a much clear condition IMHO

sergii-nosachenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinString = '';
```
sergii-nosachenko, line 8:
```suggestion
    let joinedArray = '';
```
We use verbs only for methods.

sergii-nosachenko (approved):
Good work.
Commented

PR #2116, opened 2023-02-19
added solution

Vlados3611:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let title = '';
```
sergii-nosachenko, line 8: Use a meaningful var name. It's not a title.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        title += separator;
+      };
+
+      if (this[i] !== null
+      && this[i] !== undefined) {
+        title += this[i];
+      };
```
sergii-nosachenko, line 18: Could be simplified using [Nullish coalescing operator (??)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing).

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good job 👍

PR #2117, opened 2023-02-19
Solution

ksavelyeva:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
```
sergii-nosachenko, line 8: Use a meaningful var name.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        str += this[i];
+      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+
+      if (i !== this.length - 1) {
```
```suggestion
      if (i < this.length - 1) {
```
Clearer condition

sergii-nosachenko (changes requested):

src/arrayMethodJoin.js
ksavelyeva: [solution](https://prnt.sc/MrMXLDFnhpxa)
[lint](https://prnt.sc/bmNA_9Kdg53N)
 I wrote it like that but I don't think lint like it or I have an issue which I don't get :(

sergii-nosachenko (approved):
Good work 👍

PR #2118, opened 2023-02-19
Solution

sasha-dranuta:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,21 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringJoin = '';
+    const length = this.length;
+
+    for (let i = 0; i < length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        stringJoin += this[i];
+      }
```
sergii-nosachenko, line 14: Could be simplified using [Nullish coalescing operator (??)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing).

sergii-nosachenko (approved):
Good work 👍
Left a comment for improvement.

src/arrayMethodJoin.js
sasha-dranuta: > Could be simplified using [Nullish coalescing operator (??)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing).

Ty)

PR #2122, opened 2023-02-20
array-method-join

MaksymHuriev:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
```
nazarmatsevych, line 11: You need to add separator if current index is not equal to last index

nazarmatsevych (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        result += this[i];
+      }
+
+      if (this[i] !== this[this.length - 1]) {
```
loralevitska, line 15: [Array.prototype.at()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at)
```suggestion
      if (this[i] !== this.at(-1)) {
```

loralevitska (approved):

src/arrayMethodJoin.js
MaksymHuriev: Thank you for your comments

PR #2124, opened 2023-02-20
Solution

kostiantin-hudzenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
```
IvanFesenko, line 13: Let do action if all good
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        string += this[i];
      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+
+      if (i !== 0) {
+        string += separator;
+      }
+      string += this[i];
```
IvanFesenko, line 18:
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
+      }
+
+      if (i !== 0) {
```
IvanFesenko, line 15:
```suggestion
      if (i !== this.length - 1) {
```

IvanFesenko (changes requested):
Lets do some refactor

tetiana-ivanchenko (approved):

PR #2125, opened 2023-02-20
Solution

DenysDubas:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let newString = '';
```
sergii-nosachenko, line 8: Use proper naming, f.e. `joinedArray` to describe what's inside

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        newString += this[i];
+      }
```

sergii-nosachenko (approved):

PR #2126, opened 2023-02-20
applyCustomJoin done

Stepan-TS:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinStr = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        joinStr += this[i];
+      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+
+      if (i !== this.length - 1) {
```
```suggestion
      if (i < this.length - 1) {
```
Clearer condition

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinStr = '';
```
sergii-nosachenko, line 8: Shortening is not an optimization. Also, `join` is a verb. Verbs are used for methods.

sergii-nosachenko (changes requested):

OleksiiBlezniuk (approved):
Well done!

PR #2127, opened 2023-02-20
Solution

valeriiayarzhemska:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+    let newSeparator = separator;
```
sergii-nosachenko, line 9: Useless variable

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1) {
+        newSeparator = '';
+      }
```
sergii-nosachenko, line 14: Why do you need this condition? You need just to add a separator only after elements under indexes less than `length - 1`.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+
+      if (this[i] === undefined || this[i] === null) {
+        joinedString += newSeparator;
+      } else {
+        joinedString += this[i] + String(newSeparator);
```
sergii-nosachenko, line 19: You don't need to convert the separator to string.
Adding anything to string automatically does this: https://stackoverflow.com/a/40848608

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+      }
```
sergii-nosachenko, line 20: Could be simplified using [Nullish coalescing operator (??)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing).

sergii-nosachenko (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #2128, opened 2023-02-21
join implemented

ArthurKrepko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] === 'undefined' || this[i] === null) {
+        joinedString += '';
+      } else {
+        joinedString += this[i];
+      }
```
sergii-nosachenko, line 15: The `if...else` construction can be avoided.
Could be simplified using [Nullish coalescing operator (??)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing).

sergii-nosachenko (changes requested):

Andr1yK (approved):

PR #2129, opened 2023-02-21
Solution

dudar-y:


ViktorHavryliuk (changes requested):
You can greatly reduce amount of code and checks by setting a default value to separator argument `function(separator = ',')`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,44 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
```
Viktor-Kost, line 8: it's redundant
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,44 @@
...
+          }
+          string += this[i] + separator;
+        }
+
+        for (let i = 0; i < string.length - separator.toString().length; i++) {
+          result += string[i];
+        }
+
+        return result;
+    }
```
Viktor-Kost, line 44: why do you need this switch case? 
You can pretty simplify your code like this

```
        result += this[i];
      }

      if (this[i] !== this[this.length - 1]) {
        result += separator;
      }
    }

    return result;
```

Viktor-Kost (changes requested):

Volodymyr-Melnyk-Fron10der (approved):
Cool👍

PR #2136, opened 2023-02-22
solution

serhiilysenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
nazarmatsevych, line 15:
```suggestion
      if (this[i] !== undefined && this[i] !== null) {
        result += this[i];
      }
```

nazarmatsevych (changes requested):
GJ,
just small suggestion

src/arrayMethodJoin.js
DarkMistyRoom: Please check your code again. It should work. the we skip the iteration if there is null or undefined

DarkMistyRoom (changes requested):

nazarmatsevych (approved):
👍

PR #2139, opened 2023-02-23
method-join

OlegRak86:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,37 @@
...
+      }
+    }
+
+    if (separator === undefined) {
+      for (let i = 0; i < this.length - 1; i++) {
+        str += `${this[i]},`;
+      }
+    }
+
+    this[this.length - 1] === undefined ? str = str + ''
```
YegorKorenkov, line 35: Please explain why you need this 3 lines?

YegorKorenkov (commented):
Please check this cases and fix
image

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,44 @@
...
+        if (this[i] === undefined || this[i] === null) {
+          this[i] = '';
+        }
+        str += `${this[i]},`;
+      }
+    }
+
+    if (this[this.length - 1]) {
+      str = str + this[this.length - 1];
+    }
```
IvanFesenko, line 43: To much useeless check and code duplicate.
You need one loop with check `if (this[i] !== undefined && this[i] !== null) `

IvanFesenko (changes requested):
Please refactor your code. If you have any qustions - ask them in fe_chart

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,34 @@
...
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let str = '';
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (separator !== '' && separator !== undefined) {
```
vlad-demchuk, line 15: it would be better to set default value for the argument instead of doing this check for each iteration

vlad-demchuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,38 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
nazarmatsevych, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,38 @@
...
-    // write code here
+    let str = '';
+
+    if (this.length === 0) {
+      return '';
+    }
```
nazarmatsevych, line 12: redundant, your str is empty string by default, so if there in no this.length, you will return empty string

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,38 @@
...
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (separator) {
```
nazarmatsevych, line 15: Now you have a default value for separator so this check is redundant

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,38 @@
...
+      }
+
+      if (separator === undefined) {
+        str += `${this[i]},`;
+      }
+    }
+
+    if (this[this.length - 1]) {
+      str = str + this[this.length - 1];
+    }
```
nazarmatsevych, line 37: Looks too complicated, you need only two checks:
1. `if (this[i] !== null && this[i] !== undefined)` then we add this[i] to our str
2. `if (i !== this.length - 1)` : we check if curren index is not equal to the last index,  if it's not then we add separator to our str

nazarmatsevych (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        str += `${this[i]}`;
+      }
```
Volodymyr-Melnyk-Fron10der, line 13: (Just recommendation) I would move it into variable
```suggestion
    for (let i = 0; i < this.length; i++) {
      const element = this[i];
      
      if (element !== undefined && element !== null) {
        str += `${element}`;
      }
```

Volodymyr-Melnyk-Fron10der (approved):
Well done✅

PR #2143, opened 2023-02-24
Solution

Dmytro-Kretov:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const sep = separator === undefined
+      ? ',' : separator === null ? 'null' : separator;
```
nazarmatsevych, line 9: Set a default value for separator - `','`

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
...
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let element = this[i];
+
+      if (element === null || element === undefined) {
+        element = '';
+      }
+      result += element;
```
nazarmatsevych, line 18:
```suggestion
      if (element !== null && element !== undefined) {
        result += element;
      }
```

nazarmatsevych (changes requested):
Good work,
check my comments pls

nazarmatsevych (approved):

PR #2148, opened 2023-02-28
method join

DmytroHorshunov:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,20 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        result += this[i];
+      }
+
+      if (this[i] !== this[this.length - 1]) {
```
TarasHoliuk, line 15:
```suggestion
      if (i !== this.length - 1) {
```

TarasHoliuk: check the index to know if it's the last element

TarasHoliuk (changes requested):

PR #2150, opened 2023-03-01
join() method was created

Pavlo-Petrashevskyi:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,36 @@
...
+
+        case null:
+        case undefined:
+          string += '' + `${separator}`;
+          break;
+
+        default:
+          string += this[i] + `${separator}`;
+          break;
+      }
```
OleziO, line 33: These checks are unnecessary, they degrade performance. It will be better to do it according to the following algorithm:

Each time add text and a separator to the resulting string, and at the end already check for the last element and add only the string without the separator.

OleziO (changes requested):

src/arrayMethodJoin.js
Pavlo-Petrashevskyi: done

SerhiiKirik (approved):
LGTM 🔥

PR #2151, opened 2023-03-02
complete task

VladGraund:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+      }
+
+      if (idx === this.length - 1 && el === undefined) {
+        return result;
+      }
+
+      if (el === undefined || el === null) {
+        result += `${separator}`;
+        continue;
+      }
```
IvanFesenko, line 22:
```suggestion
      if (element !== null && element !== undefined) {
        result += el;
      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+
+      if (idx < this.length - 1) {
+        result += `${el}${separator}`;
+        continue;
```
IvanFesenko, line 26:
```suggestion
        result += separator;
```

IvanFesenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (const [idx, el] of this.entries()) {
+      if (element !== null && element !== undefined) {
```
VitaliyBondarenko1982, line 11: this code does not work. element is not defined

use for loop

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #2152, opened 2023-03-03
solution

YuriiHlushenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const last = this[this.length - 1];
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += separator;
+      } else {
+        result += `${this[i]}${separator}`;
+      }
```
IvanFesenko, line 16: Also add separator only if it's not last element
```suggestion
      if (this[i] !== undefined && this[i] !== null) {
        result += this[i];
      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+    }
+
+    if (last === null || last === undefined) {
+      return result;
+    } else {
+      result += last;
+    }
```
IvanFesenko, line 23: Redudant

IvanFesenko (changes requested):
Good job. Check my comments

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+    for (let i = 0; i < len; i++) {
+      if ((this[i] === null || this[i] === undefined) && (i !== len - 1)) {
+        result += separator;
+      } else if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else if (i === len - 1) {
+        result += `${this[i]}`;
+      } else {
+        result += `${this[i]}${separator}`;
+      }
```
loralevitska, line 20: Try to optimize your solution

```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        result += this[i];
      } 
      
      if (i < len - 1) {
        result += separator;
      } 
```

loralevitska (changes requested):

etojeDenys (approved):
LGTM

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const len = this.length;
+
+    for (let i = 0; i < len; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
```
OleziO, line 12: Replace with `this[i]`, it will be the same as `this[i] !== null && this[i] !== undefined`

OleziO (changes requested):

src/arrayMethodJoin.js
YuriiHlushenko: image
but we lose 0 NaN

Anastasiia-Svintsova (approved):

PR #2153, opened 2023-03-04
Solution

Kirillets:


Volodymyr-Melnyk-Fron10der (changes requested):
you can't use Array.join() (

Kirillets:
I fix it

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,20 @@
  * Implement method join
  */
```
OleziO, line 5: Remove all comments.

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
```
OleziO, line 11: Replace with `this[i]`, it will be the same as `this[i] !== null && this[i] !== undefined`

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,20 @@
...
+        result += this[i];
+      }
+
+      if (i !== this.length - 1) {
```
OleziO, line 15: Better to use `i < this.length - 1`

OleziO (changes requested):

Kirillets:
I don`t understand what you mean 
Replace with this[i], it will be the same as this[i] !== null && this[i] !== undefined

SerhiiKirik (approved):
Well done 🔥

PR #2161, opened 2023-03-09
1st commit

ihor-halkin:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,63 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
polosanya, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```
You can set the default separator. You can remove most of checks in that case. It will make your code much more cleaner)

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,63 @@
...
+
+      return result;
+    } else {
+      for (const elem of this) {
+        if (elem === null || elem === undefined) {
+          result += ',';
+          continue;
+        }
+
+        result += elem + separator;
```
polosanya, line 60:
```suggestion
        if (elem !== null && elem !== undefined) {
          result += elem + separator;
        }
```
Just add the value to result only if it's not null or undefined

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,63 @@
...
+      }
+
+      return result.slice(0, -1);
```
polosanya, line 63: But what if separator consists of more than 1 character? 
Better use a default loop, so you can get the index and check if it's not the last iteration. Add separator only if it's true

polosanya (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (separator === null) {
+      for (const elem of this) {
+        result += elem + 'null';
+      };
+
+      return result.slice(0, -4);
```
agniya-i, line 15: this check looks redundant

agniya-i (changes requested):

nazarmatsevych (approved):
👍

PR #2163, opened 2023-03-10
solution

daryna-hnidash:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let separ = separator;
+
+    if (separator === null) {
+      separ = 'null';
+    }
```
IvanFesenko, line 12: Redudant

IvanFesenko (approved):
Try to implement using `for (let i = 0; i < this.length; i++)` loop

src/arrayMethodJoin.js
daryna-hnidash: it doesn`t work with null as a parametr

PR #2164, opened 2023-03-11
Solution

LBondarchuk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let newArray = '';
```
ivbutsykin, line 8: Let's go rename this in `result` for example. Because `newArray` name expected array as value)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+      if (this[i] !== null && this[i] !== undefined) {
+        newArray += this[i];
+      }
+
+      if (i < this.length - 1) {
+        newArray += separator;
+      }
+    }
+
+    return newArray;
```
ivbutsykin, line 20: Code easy for understanding, thanks

ivbutsykin (approved):
Good job!

Don't forget rename variable)

PR #2166, opened 2023-03-12
FedenkoArrayMethod

VladFedenko:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,32 @@
...
+
+    for (let i = 1; i < this.length; i++) {
+      result += `${newSep}`+ `${newArr[i]}`;
+    }
+
+    if (this.length === 0) {
+      result = '';
+    }
+
+    return result;
```
etojeDenys, line 33: think, how you can rewrite it using only one `for` (it's possible)

etojeDenys (changes requested):
or you always can ask for help in the slack chat

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const newArr = [];
+    let result = [];
+    let newSep = `${separator}`;
+
+    if (separator === undefined) {
+      newSep = ',';
+    }
```
VitaliyBondarenko1982, line 14: instead this checking you could use default parameter
```
function(separator === ',')

VitaliyBondarenko1982 (approved):

PR #2167, opened 2023-03-13
Solution

yaroslav5481:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        result += this[i];
+      }
+
+      if (this.length > 1 & i < this.length - 1) {
```
slavanosatskyi, line 15: redundant check, if length is 1, then i < this.length will alway be false as i can be equal to 0 only, and 0 < 1 - 1 is false
```suggestion
      if (i < this.length - 1) {
```

slavanosatskyi (changes requested):

loralevitska (approved):
👍

PR #2168, opened 2023-03-13
Solution

yao157:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,22 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
nazarmatsevych, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

You can set a default value for separator

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,22 @@
...
-    // write code here
+    const sep = separator === undefined ? ',' : separator;
```
nazarmatsevych, line 8:
```suggestion
    const sep = separator;
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,22 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        result += sep;
+      }
+
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
nazarmatsevych, line 20:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        result += this[i];
      }
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,22 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const sep = separator === undefined ? ',' : separator;
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
```
nazarmatsevych, line 12:
```suggestion
      if (i !== this.length - 1) {
```

if it's not the last index

nazarmatsevych (changes requested):

PR #2173, opened 2023-03-14
first version

Vladimir-Zadorozhnyi:


src/arrayMethodJoin.js
```diff
@@ -3,9 +3,29 @@
...
+
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    let value = separator;
+
+    if (value === undefined) {
+      value = ',';
+    }
```
anstsot, line 14: you can simplify it by adding the default value on line 8 (`separator = ','`) or:
```suggestion
    const value = separator || ',';
```

anstsot (approved):
Great job, approved!

PR #2174, opened 2023-03-14
create method join

allagutsul27:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,27 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const joinSeparator = separator;
+
+    const arr = [...this];
```
IvanFesenko, line 10: This is redudant

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,27 @@
...
+
+    let res = '';
+
+    for (let i = 0; i < arr.length; i++) {
+      if (arr[i] === null || arr[i] === undefined) {
+        arr[i] = '';
+      }
```
IvanFesenko, line 17: You should do action if value not equal null or undefined
```suggestion
      if (arr[i] !== null || arr[i] !== undefined) {
        res += arr[i];
      }
```

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,27 @@
...
+
+      if (i === arr.length - 1) {
+        res += arr[i];
+      } else {
+        res += arr[i] + '' + joinSeparator;
+      }
```
IvanFesenko, line 23: If it's not last item - you add separator
```suggestion
      if (i !== arr.length - 1) {
        res += separator;
      }
```

IvanFesenko (changes requested):
Check my comments

mykhalenych (approved):

PR #2176, opened 2023-03-14
method-join-solution

StepuraSerhii:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        result += this[i];
+      }
+
+      if (i === this.length - 1) {
+        break;
+      }
+      result += separator;
```
polosanya, line 18:
```suggestion
      if (i !== this.length - 1) {
        result += separator;
      }
```
Just add separator only if it's not the last iteration.
It will make your code a bit cleaner

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
polosanya, line 8: It's not a good idea to use "result" as a name for your variables. 
Let's try to make the name more descriptive)

polosanya (changes requested):

DarkMistyRoom (approved):

PR #2179, opened 2023-03-15
Develop

Vaskovics:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
```
OleziO, line 11: Replace with `!this[i]`, it will be the same as `this[i] === undefined || this[i] === null`

OleziO (approved):

PR #2180, opened 2023-03-16
Solution

JaneShavrukova:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    const modValue = function(value) {
+      return value === null || value === undefined ? '' : value;
+    };
+
+    for (let i = 0; i < this.length; i++) {
+      result += modValue(this[i]);
```
tetiana-ivanchenko, line 15: it's better to use if block, because now, if your value is null or undefined you're doing another redundant operation - you're adding an empty string, it's better to avoid it

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):
GJ

PR #2182, opened 2023-03-16
solution

popeye88:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let actualSep = separator;
+
+    if (separator === undefined) {
+      actualSep = ',';
+    }
```
qoa11a, line 12: Instead of doing this, you could just write separator = ',' inside function parentheses.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
qoa11a, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

qoa11a (approved):

PR #2184, opened 2023-03-16
solution method-join

glots-a:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        this[i] = '';
```
VinnieJ-2k20, line 12: Don't mutate array, original method doesn't do it.

VinnieJ-2k20 (changes requested):
Solution needs one fix.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const copyThis = [...this];
```
VinnieJ-2k20, line 9:
```suggestion
    const copyArray = [...this];
```

VinnieJ-2k20 (approved):
Good job 👍 
Not the best variable name.

PR #2188, opened 2023-03-16
function works

BondarenkoDmitriy:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    let separ = separator;
```
Stambuliak, line 9: fix variables of these two

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
...
+
+    if (separ === undefined) {
+      separ = ',';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      const verificatStr = this[i];
```
Stambuliak, line 16: sounds bad. Change variable name

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
...
+
+      if (verificatStr === null || verificatStr === undefined) {
+        result += '';
+      } else if (verificatStr.toString() === 'null') {
+        result += 'null';
+      } else {
+        result += verificatStr.toString();
```
Stambuliak, line 23: we do not need to transform to String. It has already done.

Stambuliak (changes requested):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let joinStr = '';
+    let tempSeparator = separator;
+
+    if (tempSeparator === undefined) {
+      tempSeparator = ',';
+    }
```
VinnieJ-2k20, line 13: Just add default value for function parameter.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let joinStr = '';
```
VinnieJ-2k20, line 8:
```suggestion
    let joinedString = '';
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
...
+    let tempSeparator = separator;
+
+    if (tempSeparator === undefined) {
+      tempSeparator = ',';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      const checkStr = this[i];
```
VinnieJ-2k20, line 16:
```suggestion
      const element = this[i];
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
...
+
+      if (checkStr === null || checkStr === undefined) {
+        joinStr += '';
+      } else if (checkStr === 'null') {
+        joinStr += 'null';
+      } else {
```
VinnieJ-2k20, line 22: Else if condition will always be false. You need to alter the logic.
if `checkStr === null` is true - else if will never execute.

VinnieJ-2k20 (changes requested):
You are on the right way. Some changes needed.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedStr = '';
+    const tempSeparator = separator;
```
liliia-livandovskaya, line 9:
```suggestion
```
why do you need `tempSeparator`? Use just `separator`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedStr = '';
```
```suggestion
    let joinedString = '';
```

liliia-livandovskaya (changes requested):
GJ! Lets work on improvement

Olexandr6 (approved):
Good job 👍

PR #2190, opened 2023-03-16
join method ready for use

PavloReutskiy:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,31 @@
...
+
+    if (divider === undefined) {
+      divider = ',';
+    }
+
+    let string = '' + sourceAarrayCopy[0];
+
+    for (let i = 1; i < sourceAarrayCopy.length; i++) {
+      string += `${divider}${sourceAarrayCopy[i]}`;
+    }
```
Esteban4769, line 30: Dont need first line. You can move through loop and add separators to all element except last one.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,31 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    if (this.length === 0) {
+      return '';
+    }
+
+    const sourceAarrayCopy = [...this];
```
Esteban4769, line 12: Dont need to copy array [checklist](https://github.com/mate-academy/js_array-method-join/blob/master/checklist.md) pt. 2 (main regect reason)

Esteban4769 (changes requested):
Good job, need just to make some small changes.

liliia-livandovskaya (approved):

PR #2191, opened 2023-03-17
added custom array join realization

Esteban4769:


M1k1ta (approved):
Good job! I solved the use if/ else without the ternary operator, but I think it is also a good solution.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      result += this[i] === null || this[i] === undefined
+        ? '' : this[i];
```
liliia-livandovskaya, line 12:
```suggestion
      result += this[i] === null || this[i] === undefined
        ? ''
        : this[i];
```

liliia-livandovskaya (approved):

PR #2194, opened 2023-03-17
solution

artem-kyrylenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedArray = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        joinedArray += '';
+      } else {
```
PavloReutskiy, line 13: as I remember else is not cool

PavloReutskiy (approved):
check it once more, maybe you can avoid using else

liliia-livandovskaya (approved):

PR #2195, opened 2023-03-17
Solution

Anna-Kuzmenko:


Zad0rnik (approved):
gj

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length < 1) {
+      return '';
+    }
+
+    let str = '';
```
liliia-livandovskaya, line 12: `str` is a bad name for the variable

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+
+    for (let i = 0; i < this.length; i += 1) {
+      let item = this[i];
+
+      if (item === undefined || item === null) {
+        item = '';
+      }
```
liliia-livandovskaya, line 19:
```suggestion
```
you do not need to add ''

src/arrayMethodJoin.js
Anna-Kuzmenko: fixed

src/arrayMethodJoin.js
Anna-Kuzmenko: deleted

Anna-Kuzmenko:
I forgot to commit my solution (( sorry!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length < 1) {
+      return '';
+    }
+
+    let joinedString = '';
```
liliia-livandovskaya, line 12: it would be better to move the variable declaration to the top before `if`

PR #2197, opened 2023-03-17
Create custom Join

Hamuud:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    // Who invented this task.. i want to cry
+    const copyThis = [...this];
```
roshirin, line 9: Checklist#2 suggests to avoid creating needless copies of arrays and iterate existing array

liliia-livandovskaya: agree, you don't need to copy the array here

liliia-livandovskaya (changes requested):
GJ! Lets work on improvement

PR #2199, opened 2023-03-17
Created join method

roshirin:


AlexRog01 (changes requested):
Good solution, but try to follow the fourth point of the checklist. Also, if you add a separator before a new element, you don't have to work out adding the last element.

roshirin:
4. [CODE STYLE] - avoid using nested `if` statements, we are sure that you can do it without them :)

Can't see any nested 'if' statements in my solution. Did I miss it?

If I add a separator before a new element, I'll have to work out the first element separately. No advantage here in my opinion.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+      joinedElements += separator;
+    }
+
+    const lastElement = this[this.length - 1];
+
+    if (lastElement === null || lastElement === undefined) {
+      return joinedElements;
+    }
+
+    return joinedElements + lastElement;
```
liliia-livandovskaya, line 24: this can be simplified, try to rewrite the code

liliia-livandovskaya (changes requested):
GJ! Lets work on improvement

Olexandr6 (approved):

PR #2203, opened 2023-03-18
Solution

danylo-dobrovolskyi:


VenomMB (approved):
GJ!!!

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
liliia-livandovskaya, line 12: try to avoid nesting, you can use a conditional (ternary) operator here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator

liliia-livandovskaya (approved):

PR #2204, opened 2023-03-18
create function CustomJoin

VenomMB:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,29 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        joinedString += '';
+      } else {
+        joinedString += this[i];
+      }
```
MylaCrystal, line 15: try to rewrite this part a bit, I believe you could write a little less code here

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,29 @@
...
+    return joinedString;
   };
 }
 
+const source = [0, 1, 2, 3];
+
+applyCustomJoin();
+
+source.join2(null);
+
```
MylaCrystal, line 31: do we need this part for tests? 

(P.S. in a real project, we would probably write this data, but in the case of tasks, we will receive several different variations instead of this single option)

MylaCrystal (approved):

src/arrayMethodJoin.js
liliia-livandovskaya: agree, try to rewrite it

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,29 @@
...
+    return joinedString;
   };
 }
 
+const source = [0, 1, 2, 3];
+
+applyCustomJoin();
+
+source.join2(null);
```
liliia-livandovskaya, line 30:
```suggestion
```
you dont need it

liliia-livandovskaya (changes requested):
GJ! Lets improve your code

liliia-livandovskaya (approved):

PR #2205, opened 2023-03-18
done

victor-buldenko:


src/arrayMethodJoin.js
```diff
@@ -1,11 +1,36 @@
...
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    const sep = String(separator);
+
+    for (let i = 0; i < this.length; i++) {
+      switch (true) {
+        case (i + 1) !== this.length && this[i] === undefined:
+        case (i + 1) !== this.length && this[i] === null:
+        case (i + 1) !== this.length && this[i] === NaN:
```
TarasHoliuk, line 17: `NaN` should be just conversed to a string
image

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,36 @@
...
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    const sep = String(separator);
+
+    for (let i = 0; i < this.length; i++) {
+      switch (true) {
+        case (i + 1) !== this.length && this[i] === undefined:
```
TarasHoliuk, line 15: no need to check if it's the last element every time

instead of doing it, you may just add `this[i]` or an empty string if it's `null` or `undefined`.
and then in another if statement add a separator if it's not the last element

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,36 @@
...
+        case (i + 1) !== this.length && this[i] === null:
+        case (i + 1) !== this.length && this[i] === NaN:
+          result += ',';
+          break;
+        case this[i] === undefined:
+          break;
+        case (i + 1) !== this.length && separator === undefined:
```
TarasHoliuk, line 22: no need to check `separator === undefined` - just set the default value in function arguments

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,36 @@
...
  * Implement method join
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    const sep = String(separator);
+
+    for (let i = 0; i < this.length; i++) {
+      switch (true) {
```
TarasHoliuk, line 14: What's the matter to use `switch(true)`? Use `if` blocks instead

TarasHoliuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+        case this[i] === undefined:
+          result += sep;
+          break;
+        case (i + 1) !== this.length:
+          result += this[i] + sep;
+          break;
+        default:
+          result += this[i];
+      }
+    }
```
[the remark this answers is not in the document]
SanyaBratashchuk: It will be much more readable and look much more prettier with `if's` statements instead of `switch(true)`

SanyaBratashchuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const sep = String(separator);
```
DarkMistyRoom, line 9: This line is redundant, because you will add it to the result which is `string`, so it always will be converted to a `string`.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (i === (this.length - 1) && !this[i]) {
+        return result;
+      } else if (this[i] === null || this[i] === undefined) {
+        result += sep;
+      } else if ((i + 1) !== this.length) {
+        result += this[i] + sep;
+      } else {
+        result += this[i];
+      }
```
DarkMistyRoom, line 20: try to avoid `if else` constructions where can be used just flat `if`s

DarkMistyRoom: simplify this query chain. you only need two checks: 
- null & undefined
- should we add the separator

DarkMistyRoom (changes requested):

sTorba24 (approved):
GJ 🚀

PR #2208, opened 2023-03-18
Solution

Zibi95:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result = this.length - 1 === i
+          ? result + ''
+          : result + '' + joinOperator;
+      } else {
+        result = this.length - 1 === i
+          ? result + this[i]
+          : result + this[i] + joinOperator;
```
Radoslaw-Czerniawski, line 23: Like in my other review, when adding strings I suggest using template literals for the readability sake and to omit such situations:
```
3 * 5 => 15
"3" * "5" => 15
2 + " fish" => "2 fish"
"3" + "5" => "35"  // hmmm...
```

Radoslaw-Czerniawski (approved):
Tried it out, works like the original join. Good job :D

PR #2214, opened 2023-03-19
add solution for custom join

Pozdnya:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringFromArray = '';
+
+    if (this.length === 0) {
+      return '';
+    }
+
```
VladislavaT, line 13: if u do this check u can cover for loop in this and return string once


if (this.length  !== 0) {
      for loop here...
    }
    
    return stringFromArray;

VladislavaT (approved):

src/arrayMethodJoin.js
Pozdnya: agree. I'll change my code. Thanks.

liliia-livandovskaya (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringFromArray = '';
+
+    if (this.length !== 0) {
+      for (let i = 0; i < this.length; i++) {
+        const stringItem = this[i];
+
+        stringFromArray += stringItem === undefined || stringItem === null
+          ? '' : stringItem;
```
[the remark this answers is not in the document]
Pozdnya: Done

PR #2217, opened 2023-03-19
join solution

VladislavaT:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultStr = '';
+
+    if (this.length !== 0) {
```
kukla1989, line 10: 4. [CODE STYLE] - avoid using nested `if` statements, we are sure that you can do it without them :)
you can do: 
if (this.length !== 0) {
  return resultStr;
}

kukla1989 (approved):
I am so inspired by your solution that will rewrite mine :)

liliia-livandovskaya (approved):

PR #2218, opened 2023-03-19
solution

Evgenb000:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultingString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        resultingString += '';
+      } else {
+        resultingString += this[i];
+      }
```
vksazonov, line 15: I think you could make it shorter by checking if this[i] !== undefined...
In that way there is no need to use else

vksazonov (approved):

src/arrayMethodJoin.js
liliia-livandovskaya: agree, try to avoid nesting, you can rewrite the condition in `if` or use a conditional (ternary) operator here https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator

liliia-livandovskaya (approved):

PR #2220, opened 2023-03-19
Solution

Kavon-Skela:


TanyaS133 (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+      if (i === this.length - 1) {
+        joinedStr += this[i].toString();
+        continue;
+      }
+
+      if (this[i] === null || this[i] === undefined) {
+        joinedStr += '' + separator;
+      } else {
+        joinedStr += this[i].toString() + separator;
+      }
```
liliia-livandovskaya, line 26: there are a lot of conditions that can be simplified, try to rewrite it

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedStr = '';
```
```suggestion
    let joinedString = '';
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+        : '';
+
+      if (i === this.length - 1) {
+        joinedStr += checker;
+        continue;
+      }
+
+      if (this[i] === null || this[i] === undefined) {
+        joinedStr += checker + separator;
+        continue;
```
liliia-livandovskaya, line 22:
```suggestion
        joinedStr += checker;
      }

      if (this[i] === null || this[i] === undefined) {
        joinedStr += checker + separator;
```
why do you need `continue` here?

liliia-livandovskaya (approved):

src/arrayMethodJoin.js
Kavon-Skela: Because I have an expression of string concatenation at the end of the loop. Without the continue operator, string concatenation will be done two times in a row with if cases

PR #2221, opened 2023-03-19
Solution

DanielSorokowski:


solaryasha (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let concatening = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this.length - i === 1 && this[i] === undefined) {
+        concatening += ``;
```
solaryasha, line 12: Maybe, let's use `continue` keyword here?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+      } else if ((this[i] === null || this[i] === undefined)) {
```
solaryasha, line 13:
```suggestion
      } else if (this[i] === null || this[i] === undefined) {
```

solaryasha (changes requested):

solaryasha (approved):

PR #2223, opened 2023-03-20
solve task with function applyCustomJoin

kukla1989:


src/arrayMethodJoin.js
```diff
@@ -7,24 +7,16 @@ function applyCustomJoin() {
   [].__proto__.join2 = function(separator = ',') {
     let res = '';
```
Nazarii98, line 8: You should add full names for your values, it's helping your code to be understandable.

src/arrayMethodJoin.js
```diff
@@ -7,24 +7,16 @@ function applyCustomJoin() {
...
 
-    for (let el of this) {
-      if (el === null || el === undefined) {
-        el = '';
-      }
-
-      res += el + '' + separator;
-    }
+    for (const el of this) {
```
Nazarii98, line 10: Also here (el => element)

Nazarii98 (changes requested):
Good solution! 
Look to comments, to improve some moments.

PR #2232, opened 2023-03-20
joinMethod

ymys1:


Rage1502 (approved):

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,23 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
-  };
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
```
liliia-livandovskaya, line 13: it's better to avoid nesting, you can use a conditional operator as an option https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator

liliia-livandovskaya (approved):

PR #2234, opened 2023-03-20
alternative way to make join array function

ofenchuk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let result = '';
+    let separ = separator;
+
+    if (separator === null) {
+      separ = 'null';
+    }
```
Olexandr6, line 14: I can't understand why you need it
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      let wordToAdd = this[i];
+
+      if (this[i] === null || this[i] === undefined) {
+        wordToAdd = '';
+      }
```
Olexandr6, line 21: Let's think in the opposite way:
if not null or undefined - you should add word + separator, if not - do nothing

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+
+      if (i + 1 === this.length) {
+        return (result += wordToAdd);
+      }
```
Olexandr6, line 25: Let's add a separator only if it's not the last index. Let's delete `return` here

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let result = '';
```
Olexandr6, line 9:
```suggestion
    let joinedString = '';
```

G-YaSha (changes requested):
Almost done👍 Now, let's work on improvements:

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
```
```suggestion
```

liliia-livandovskaya (approved):

PR #2235, opened 2023-03-20
Solution

NadiiaR:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    if (this.length > 0) {
+      for (let key = 0; key < this.length; key++) {
```
tetiana-ivanchenko, line 11:
```suggestion
      for (let i = 0; i < this.length; i++) {
```

tetiana-ivanchenko: more meanfull naming, because in array we have indexes

tetiana-ivanchenko (approved):

PR #2242, opened 2023-03-21
solution

mxwolfy:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = [];
```
IvanFesenko, line 8: Join shoul return a `string`, why initial value is `array`?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+
+    if (this.length === 0) {
+      return '';
+    }
```
IvanFesenko, line 12: Useless

IvanFesenko (changes requested):

PR #2250, opened 2023-03-21
first try

Nikolaymelni:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
polosanya, line 8: It is better to use more meaningful names than just `result`)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      
+      if (this[i] !== undefined && this[i] !== null) {
```
polosanya, line 12:
```suggestion
    for (let i = 0; i < this.length; i++) {
      if (this[i] !== undefined && this[i] !== null) {
```

polosanya (approved):
Good job!

PR #2251, opened 2023-03-22
solution

maksym-kostetskyi:


loralevitska (changes requested):
Almost done) Let's improve

IvanFesenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
polosanya, line 8: It is better to avoid naming variables "result". Try to provide more meaningful names

polosanya (approved):
Good job!

PR #2252, opened 2023-03-22
develop

nata-prog123:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,36 @@
...
+      if ((this[this.length - 1] !== null)
+       && (this[this.length - 1] !== undefined)) {
+        newProto += `${this[this.length - 1]}`;
+      }
+    } else {
+      for (let i = 0; i <= this.length - 2; i++) {
+        newProto += `${this[i]}` + ',';
+      }
+      newProto += `${this[this.length - 1]}`;
+    }
```
tetiana-ivanchenko, line 35: youe algorithm is very difficult, let's simplify it
actually you need just one loop where you will have two checks:
1. if your current element isn't null and isn't undefined, you will add it to the resulted string
2. if your element isn't last, you will add separator to the resulted string

that's it

tetiana-ivanchenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,36 @@
...
+    let newProto = '';
+    let separatorNew = separator;
+
+    if (this.length === 0) {
+      return newProto;
+    }
+
+    if (separatorNew === undefined) {
+      separatorNew = ',';
+    }
```
etojeDenys, line 17: redundant

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,36 @@
...
+
+      if ((item === null) || (item === undefined)) {
+        item = '';
+      }
+
+      if (i !== this.length - 1) {
+        newProto += item + separatorNew;
+      } else {
+        newProto += item;
+      }
```
etojeDenys, line 34:
```suggestion
      if ((item === null) && (item === undefined)) {
        newProto += item;
      }

      if (i !== this.length - 1) {
        newProto += separatorNew;
      } 
```

etojeDenys (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+      if ((item === null) || (item === undefined)) {
+        item = '';
+      }
+
+      newProto += item;
+
+      if (i !== this.length - 1) {
+        newProto += separatorNew;
+      }
+    }
```
Viktor-Kost, line 31:
```suggestion
    for (let i = 0; i < this.length ; i++) {
      let item = this[i];
      if ((item === null) || (item === undefined)) {
        joinedString += item;
      }

      if (i < this.length - 1) {
        joinedString += separator;
      }
    }
```

Viktor-Kost (changes requested):
Nice!

So, you can simplify your solution

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let item = this[i];
+
+      if ((item === null) || (item === undefined)) {
```
OleziO, line 13: Replace with `!item`, and it will be the same.

OleziO (approved):

PR #2253, opened 2023-03-22
made join method

bohdanmayor:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const item = this[i];
+
+      if (typeof item === 'string'
+        || typeof item === 'number'
+        || typeof item === 'boolean') {
```
OlEzhA0, line 16: you can check your item only for null or undefined, because we can pass even function here.
Screenshot 2023-03-23 at 9 35 19 AM

```suggestion
      if (item !== null && item !== undefined) {
        str += item
      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+        str += '' + separator + item;
+      } else {
+        str += '' + separator;
```
OlEzhA0, line 19: you can check
```
if (i < this.length - 1) {
  str += separator
}
```

OlEzhA0 (changes requested):

src/arrayMethodJoin.js
bohdanmayor: it s have no sence

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const item = this[i];
+
+      if (item !== null && item !== undefined) {
+        str += '' + separator + item;
```
DarkMistyRoom, line 15: why do we need an empty string here if `str `is already a string

DarkMistyRoom: then what about the first element? why do you add the separator before it?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+      } else {
+        str += separator;
+      }
+    }
+
+    if (separator !== '') {
+      const sep = `${separator}`;
+
+      str = str.substring(sep.length);
+    }
```
DarkMistyRoom, line 25: let's change the logic a bit to make the code simpler. two plane ifs will be enough. decide if we need to add an element if the first one, and if we need to add the separator in the second

src/arrayMethodJoin.js
DarkMistyRoom: why do you think so? if you have questions, ask in the chat

DarkMistyRoom (changes requested):
Let's improve the solution

src/arrayMethodJoin.js
bohdanmayor: firstly we add separator before element, and then we are delete it on first position

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+      if (item !== null && item !== undefined) {
+        if (i !== 0) {
+          str += newSeparator + item;
+        } else {
+          str += item;
+        }
+      } else {
+        if (i !== 0) {
+          str += newSeparator;
+        }
```
VitaliyBondarenko1982, line 24:
```suggestion
  if (item !== null && item !== undefined) {
        str += i ? newSeparator + item : item;
      } else if (i !== 0) {
        str += newSeparator;
      }
```

VitaliyBondarenko1982 (approved):

PR #2258, opened 2023-03-28
solution

Fedos17:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+      return '';
+    }
+
+    if (res === null) {
+      res = '';
+    }
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        res += String(separator) + '';
```
DarkMistyRoom, line 20:
```suggestion
        res += separator;
```

DarkMistyRoom: `res `is already a string

DarkMistyRoom: check the types transformation rules

DarkMistyRoom (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let res = this[0];
+
+    if (!this.length) {
+      return '';
+    }
+
+    if (res === null) {
+      res = '';
+    }
```
IvanFesenko, line 16: This is redudant if fix `if` condition
```suggestion
    let res = '';
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        res += separator + '';
+        continue;
+      }
+      res += String(separator) + this[i];
```
IvanFesenko, line 23: Also you should separator if it's not the last element
```suggestion
      if (this[i] !== null || this[i] !== undefined) {
        res += this[i];
      }
      
      if (i < this.length - 1) {
        result += separator;
      }
     
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+    }
+
+    res = String(res);
```
IvanFesenko, line 26: It's already string
```suggestion
```

IvanFesenko (changes requested):

PR #2261, opened 2023-03-29
solution

MikiOOOO:


Dka888 (approved):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let finalString = '';
+    const finalSeparator = separator === undefined ? ',' : '' + separator;
```
d-wasilewski: maybe add a default separator value instead of checking if it's undefined

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let finalString = '';
```
d-wasilewski: be more descriptive, finalString doesnt say much

d-wasilewski (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+      currentElement = this[i];
+
+      if (currentElement === null || currentElement === undefined) {
+        currentElement = '';
+      }
+
+      if (i === this.length - 1) {
+        linkedStringParts += currentElement;
+      } else {
+        linkedStringParts += currentElement + '' + separator;
```
d-wasilewski: is this empty string necessary here? Also, you can use template string

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let linkedStringParts = '';
+    let currentElement;
+
+    for (let i = 0; i < this.length; i++) {
+      currentElement = this[i];
+
+      if (currentElement === null || currentElement === undefined) {
+        currentElement = '';
```
d-wasilewski: maybe set '' as a default value and just continue the loop here instead of reassigning it

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2262, opened 2023-03-29
first solution

spencerspe:


polosanya (changes requested):
Almost done! Let's remake your solution without using switch...case. It's actually redundant in this task

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,30 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let newString = '';
+
+    if (this.length === '0') {
```
anstsot, line 11:
```suggestion
    if (this.length === 0) {
```

anstsot (approved):
Good job, approved!

PR #2265, opened 2023-03-31
Solution

HannaVasylieva:


witflash (changes requested):
Nice solution.
But:
image

Could you try to do this task without `toString` method?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let connectedParts = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] != null) {
+        connectedParts += this[i];
+      }
```
IvanFesenko, line 13: Don't forget about undefined
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        connectedParts += this[i];
      }
```

IvanFesenko (approved):

PR #2267, opened 2023-04-01
Solution

mateusz-michalowski:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,24 @@
...
+    let joinedStr = '';
+    const arrLength = this.length;
+
+    for (let i = 0; i < arrLength; i++) {
+      const type = typeof (this[i]);
+
+      if (type === 'number' || type === 'string' || type === 'boolean') {
+        joinedStr += this[i];
+      }
+
```
sylcym, line 16: I think it will be enough to check if `this[i]` is `null `or `undefined`, if so skip the iteration to the next using `continue`

sylcym (changes requested):

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,22 @@
...
     // write code here
+    let joinedStr = '';
+    const arrLength = this.length;
+
+    for (let i = 0; i < arrLength; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        joinedStr += this[i];
+      }
+
+      if (typeof (separator) !== 'undefined' && i < arrLength - 1) {
```
d-wasilewski: you don't have to use `typeof` here, maybe try setting the default separator argument value instead of checking it?

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2268, opened 2023-04-02
solution

ghostargg:


kacper-lyczba (approved):

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,31 @@
...
   [].__proto__.join2 = function(separator) {
     // write code here
+    if (this.length === 0) {
+      return '';
+    }
+
+    let separateSign = separator;
+
+    if (separateSign === undefined) {
+      separateSign = ',';
```
d-wasilewski: in this case you can just assign it as a default value

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,31 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
     // write code here
+    if (this.length === 0) {
+      return '';
+    }
+
+    let separateSign = separator;
```
d-wasilewski: is this really necessary?

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,31 @@
...
+
+    if (separateSign === undefined) {
+      separateSign = ',';
+    } else if (separateSign === null) {
+      separateSign = 'null';
```
d-wasilewski: i don't think we want to have null in the output string

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,31 @@
...
+    }
+
+    let returnedValue = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        returnedValue += (i === this.length - 1) ? '' : separateSign;
```
d-wasilewski: why do you need to check this condition `(i === this.length - 1` here?

d-wasilewski (changes requested):

src/arrayMethodJoin.js
d-wasilewski: I get it, it's fine

src/arrayMethodJoin.js
ghostargg: i get this error if i do not create variable for parameter 'Assignment to function parameter 'separator''

src/arrayMethodJoin.js
ghostargg: there is test that says our function should work with null as parameter

src/arrayMethodJoin.js
ghostargg: im checking if this is the last element of array if it is i dont add any separator sign

d-wasilewski (approved):

PR #2269, opened 2023-04-02
Solution

sylcym:


YBM3 (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        string = string + separator;
```
d-wasilewski: checklist 3

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+      }
+
+      if (this[i] === null || this[i] === undefined) {
+        continue;
+      } else {
+        string = string + this[i];
+      }
+    }
```
d-wasilewski: the use of an if...else statement with a continue statement can be simplified. You can use an if statement without an else block to only execute the code if the condition is true

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
```
d-wasilewski: use more descriptive name

d-wasilewski (changes requested):

src/arrayMethodJoin.js
sylcym: done 😅

src/arrayMethodJoin.js
sylcym: done

src/arrayMethodJoin.js
sylcym: updated

d-wasilewski (approved):

PR #2271, opened 2023-04-02
solution v1

bjeziorek:


Ozwiena (approved):

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,57 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
     // write code here
+    let safeSeparator = '';
```
d-wasilewski: just use default separator value

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,57 @@
...
+    }
+
+    for (let i = 0; i < this.length - 1; i++) {
+      joinedString += ''
+        + replaceNullAndUndefinedWithEmptyString(this[i])
+        + '' + (safeSeparator || '');
+    }
+
+    joinedString += ''
+      + replaceNullAndUndefinedWithEmptyString(this[this.length - 1]);
```
d-wasilewski: it's not really readable this way. Try to do regular checks here instead of this funciton. Are all of them really necessary? Try to simplify your code

d-wasilewski (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,39 @@
...
+  [].__proto__.join2 = function (separator = ',') {
+    let joinedString = '';
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    if (this.length === 0) {
+      return [];
+    }
```
d-wasilewski: they check for the same thing

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,39 @@
...
+
+    if (this.length === 1) {
+      return '' + this[0];
+    }
+
+    if (!this[1] && this[0]) {
```
d-wasilewski: a check for !this[1] && this[0] that seems unnecessary and unclear in its intent

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,39 @@
...
+      return this;
+    }
+
+    for (let i = 0; i < this.length - 1; i++) {
+      joinedString += ''
+        + ((this[i] === undefined || this[i] === null)
+          ? ''
+          : this[i])
+        + '' + ((separator === null ? 'null' : separator) || '');
```
d-wasilewski: try to use if statements to make it more readable and clear please

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,39 @@
...
+    }
+
+    joinedString += ''
+      + ((this[this.length - 1] === undefined || this[this.length - 1] === null)
```
d-wasilewski: do you need both of these checks here?

d-wasilewski (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,42 @@
...
+      }
+
+      if (separator === null) {
+        joinedString += 'null';
+      }
+
+      if (separator !== null) {
+        joinedString += separator;
+      }
+    }
```
d-wasilewski: you can use template string instead of checking for null and assigning the string `null` to joinedString. Template string will convert null to 'null' on its own

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,42 @@
...
+      return '';
+    }
+
+    if (this.length === 1) {
+      return '' + this[0];
+    }
+
+    if (this[1] === undefined) {
+      return this;
+    }
```
d-wasilewski: can you try to do it without these check aswell? Just delete them and play with the conditions below

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2274, opened 2023-04-03
Solution

AnastasiiaVorobets:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        result += separator;
+      }
+
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
IvanFesenko: You should add separator at the end in case when it's not the last element
No need do action if `this[i]` is equal to `null` or `undefined`
```suggestion
      if (this[i] !== null || this[i] !== undefined) {
        result += this[i];
      } 
      
      if (i !== this.length - 1) {
        result += separator;
      }
```

IvanFesenko (changes requested):

nazarmatsevych (approved):

PR #2276, opened 2023-04-03
solution

Ozwiena:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedArray = '';
+
+    if (this[0] !== null) {
+      joinedArray += this[0];
+    }
+
+    let tool = separator;
```
d-wasilewski: it's not making it more readable

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedArray = '';
+
+    if (this[0] !== null) {
+      joinedArray += this[0];
+    }
```
d-wasilewski: is this check necessary?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+
+    let tool = separator;
+
+    if (tool === null) {
+      tool = 'null';
+    }
```
d-wasilewski: this way we will have nullls in the string, I don't think we want that

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+      if (this[i] !== undefined && this[i] !== null) {
+        joinedArray += tool + this[i];
+      } else {
+        joinedArray += tool + '';
+      }
+    }
+
+    if (this.length === 0) {
+      joinedArray = '';
+    }
```
d-wasilewski: this should be in the beginning of the function, if the length is 0 you can just return there

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+
+    let tool = separator;
+
+    if (tool === null) {
+      tool = 'null';
+    }
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        joinedArray += tool + this[i];
```
d-wasilewski: shouldn't you start with an element and then append the separator?

d-wasilewski (changes requested):

src/arrayMethodJoin.js
Ozwiena: I think we need nulls in the string for one of the tests:

test(`Should work with null`, () => {
  expect(source.join2(null))
    .toBe('0null1null2null3');
});

src/arrayMethodJoin.js
Ozwiena: Do you mean the name of the variable?

src/arrayMethodJoin.js
Ozwiena: When I remove this check, two tests are failed:

1)  test(`Should work when array has null as a first element`, () => {
      78 |   expect([null, null, 1, undefined, 0, NaN, ''].join2(','))
    > 79 |     .toBe(',,1,,0,NaN,');
         |      ^
         
2) test(`Should work when array contains a string 'null'`, () => {
      83 |   expect([null, null, 1, undefined, 0, 'null', NaN, ''].join2(','))

src/arrayMethodJoin.js
Ozwiena: My initial joinedArray already has an element (this[0]). Is it really necessary to change the code here? :)

src/arrayMethodJoin.js
d-wasilewski: I mean assignment this separator to a new variable, you should just use it directly

src/arrayMethodJoin.js
d-wasilewski: It's true, but you should use template string to join the separator and the element, it will convert null to string by itself, you won't need to do it manually. Due to that you won't need to assign the separator variable to tool variable (prev comment)

src/arrayMethodJoin.js
d-wasilewski: My point is that with this change you probably won't need to have this[0] in the array initially, which will make the code cleaner

src/arrayMethodJoin.js
d-wasilewski: > My point is that with this change you probably won't need to have this[0] in the array initially, which will make the code cleaner

related to this

d-wasilewski (changes requested):

src/arrayMethodJoin.js
Ozwiena: OK, I get it, I will try

d-wasilewski (approved):

PR #2277, opened 2023-04-03
add solution

PufAnimalsDev:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+
+    let joinedExpression = '';
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        joinedExpression += separator;
+        continue;
+      }
+      joinedExpression += this[i];
+      joinedExpression += separator;
```
d-wasilewski: you check if its null but even if it is you still add it after the is statement, I don't think it's intended

d-wasilewski (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+    if (this.length === 0) {
+      return '';
+    }
+
+    let joinedExpression = '';
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        joinedExpression += this[i];
+        joinedExpression += separator;
```
d-wasilewski: You can do this in one line with template string

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2283, opened 2023-04-04
solution

KielpinskiJ:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
d-wasilewski: Be more descriptive. What is the result?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        if (i !== this.length - 1) {
+          result += String(separator);
+        }
+      } else if (i !== this.length - 1) {
+        result += this[i] + String(separator);
```
d-wasilewski: you dont have to parse it to string I think

d-wasilewski (changes requested):
check checklist point 4

d-wasilewski (approved):

PR #2286, opened 2023-04-04
solve

ctgdprzyb:


d-wasilewski (changes requested):
1. he code creates a new array mergedArray and loops over the original array to copy its values into it. This is unnecessary since you can loop over the original array directly (checklist)
2. The code uses both string and result to refer to the same variable, which can be confusing

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+    let result = '';
+    let firstLoop = true;
+
+    for (let string of this) {
+      if (string === undefined || string === null) {
+        string = '';
+      }
+
+      if (firstLoop) {
+        result = String(string);
```
d-wasilewski, line 17: Do you need to parse this to string here? If so, use template string, if not just remove it

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+        firstLoop = false;
+      } else {
+        result = `${result}${separator}${string}`;
```
d-wasilewski, line 20: won't this do the thing?
```suggestion
        result += `${separator}${string}`;
```

d-wasilewski (changes requested):

PR #2287, opened 2023-04-04
solution

NadieinOleh:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,23 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
polosanya, line 15:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        result += this[i];
      }
```
Just add element only if it's not an empty value

polosanya (approved):
Nice!

PR #2289, opened 2023-04-04
The implementation of the join method is complete

hyrych-max:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,27 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    let delimiter = separator;
+    const copyArr = [...this];
+
+    if (delimiter === 0 || delimiter === undefined) {
+      delimiter = ',';
```
vlad-demchuk: we can just set a default value for this argument

vlad-demchuk (changes requested):

anstsot (approved):
Great job, approved!

PR #2290, opened 2023-04-05
arrayMethod-Join

theFallcon:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultStr = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (!Object.is(this[i], null) && !Object.is(this[i], undefined)) {
```
polosanya: Consider using strict equality operator `===` instead of `Object.is()` for better readability and consistency. 🤓

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultStr = '';
```
polosanya: I think it's better to use a more descriptive variable name instead of `resultStr`. For example, `joinedArray`

polosanya (changes requested):
Good job! Let's improve your code a bit)

agniya-i (approved):

PR #2291, opened 2023-04-06
solution js_array-method-join

stnayn:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1 && (this[i] == null || this[i] === undefined)) {
+        str += '';
+      } else if (i === this.length - 1) {
+        str += this[i];
+      } else if (this[i] == null || this[i] === undefined) {
+        str += separator;
+      } else {
+        str = str + this[i] + separator;
+      }
```
sTorba24: i propose you to improve it
if element !-- undefined or null => add element
if it's not last element => add separator
2 ifs will be enough for this method :)

sTorba24 (changes requested):
Well done! Little change to do :)

sTorba24 (changes requested):
fix above

sTorba24 (approved):
Much better :) Approved!

PR #2295, opened 2023-04-09
Solution

MaksymKos:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    let newSeperator = separator;
+
+    if (separator === undefined) {
+      newSeperator = ',';
+    }
```
```suggestion
  
```
instead set ` ','` as default function argument

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    let newSeperator = separator;
```
```suggestion
    const newSeparator = String(separator);
```
after  fix prev comment about default argument change let on const and fix typo

VitaliyBondarenko1982 (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
```
VitaliyBondarenko1982, line 10: no need to remove this line
```suggestion
 
 const newSeparator = String(separator);
```

VitaliyBondarenko1982: and use newSeparator in loop

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #2298, opened 2023-04-10
array-method-join

IvanOrinchak:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let join = '';
```
witflash: It's better to name it `resultString` for example or smth like that.
Because `join` mostly uses as a verb for function names.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        join += this[i];
+
+        join += i !== this.length - 1
+          && this.length - 1 !== undefined
+          ? separator : '';
+      } else if (i !== this.length - 1 && this.length - 1 !== undefined) {
+        join += separator;
+      }
```
witflash: You can do it simpler:
```
if (...) {
  resultString += this[i];
}

if (i < this.length) {
  resultString += separator;
}
```

witflash (changes requested):
Good work!
But let's try to improve your code.

agniya-i (approved):

PR #2299, opened 2023-04-11
solution

mariuszgit:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let resultString = '';
+    let tempSeparator = separator;
+
+    if (tempSeparator === undefined) {
+      tempSeparator = ',';
+    }
```
DorotaLeniecDev: Try using default value for this parameter in function definition.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let resultString = '';
+    let tempSeparator = separator;
```
DorotaLeniecDev: Do you need this tempSeparator?

DorotaLeniecDev (changes requested):
Good solution but you can improve it a bit. Check my comments

Radoslaw-Czerniawski (approved):
Very good implementation :D, for the future try to use template strings rather than adding them as javascript sometimes does some crazy things like:
```
3 * 5 => 15
"3" * "5" => 15
2 + " fish" => "2 fish"
"3" + "5" => "35"  // hmmm...
```

PR #2303, opened 2023-04-13
Solution

yura-enttsel:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
```
OleziO: Replace with `i`, it will be the same as `i > 0`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+        resultString += separator;
+      }
+
+      if (this[i] === null || this[i] === undefined) {
```
OleziO: Replace `this[i] === null || this[i] === undefined` with `!this[i]`

OleziO (changes requested):

nazarmatsevych (approved):
👍

PR #2305, opened 2023-04-13
Solution

SergeyGeorg:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,35 @@
...
+        case null:
+          arr[i] = '';
+          break;
+      }
+
+      if (i === 0) {
+        str = str + arr[i];
+      } else {
+        str = str + separator + arr[i];
+      }
```
loralevitska: You need only 2 cases
1. arr[i] !== undefined && arr[i] !== null - to exclude it from array
2. i !== this.length - 1 - to disable separator adding after last element

That's all
Improve your solution and remove redundant code

loralevitska (changes requested):
Keep up the good work! Some comments to fix:

src/arrayMethodJoin.js
SergeyGeorg: 2, I don't use after I use before delimiter

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    const arr = [...this];
```
polosanya: Creating a new array 'arr' is unnecessary. You can directly use 'this' to access the original array and avoid the extra copy. 🤔

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+
+    for (let i = 0; i < arr.length; i++) {
+      switch (arr[i]) {
+        case undefined:
+          arr[i] = '';
+          break;
+        case null:
+          arr[i] = '';
+          break;
+      }
```
polosanya: Using a switch statement for only two cases is not optimal. Consider using an if statement instead. Just check if element is not undefined or null and add it to the str only in that case

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+
+      if (i === 0) {
+        str = str + arr[i];
+      } else {
+        str = str + separator + arr[i];
```
polosanya: Think about optimizing this part too

polosanya (changes requested):
Almost done! Let's improve your code a bit)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
  * Implement method join
```
OleziO, line 4: Remove all comments.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let tmp = this[i];
+
+      if (tmp === undefined || tmp === null) {
```
OleziO, line 14: Replace `tmp === undefined || tmp === null` with `!tmp`

OleziO (changes requested):

src/arrayMethodJoin.js
SergeyGeorg: if  chenge !tmp
undefined = null = NaN = 0

test(`Should work when array has null as a first element`, () => {
  expect([null, null, 1, undefined, 0, NaN, ''].join2(','))
    .toBe(',,1,,0,NaN,');    result=  .toBe(',,1,,,,');

agniya-i (approved):

PR #2306, opened 2023-04-13
Join Method Arr OkM

Oksana-Maruniak:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
```
SerhiiKirik, line 8: Try to give more clearly name for your variables
```suggestion
    let newString = '';
```

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,21 @@
...
+
+    for (const el of this) {
```
SerhiiKirik, line 10:
```suggestion
    for (const element of this) {
```

SerhiiKirik (approved):
Well done 🔥

PR #2307, opened 2023-04-14
Solution

VladyslavKucher:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (this.length === 0) {
+      return '';
+    }
```
witflash: You can avoid this checking. If you array will be empty, your cycle will be skipped and function returns initial result with an empty string.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
```
witflash: You can do it vice versa. Check if `this[i] !== null && this[i] !== undefined` and add to result if only true.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+        result += '';
+      } else {
```
witflash: Then this `else` you will not need.

witflash (changes requested):
Good work 👍🏻
Just make a few improvements, please.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
polosanya, line 8: Better use more meaningful name for variable than just "result")

polosanya (approved):
Good job!

PR #2310, opened 2023-04-17
implemented method join

vpdrabynko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        result += separator;
+      }
+
+      if (this[i] !== null && this[i] !== undefined) {
+        result += this[i];
+      }
+    }
```
IvanFesenko: You should add separator after adding value and excluding only last element
```suggestion
    for (let i = 0; i < this.length; i++) {
      if (this[i] !== null && this[i] !== undefined) {
        result += this[i];
      }
      
      if (i < this.length - 1) {
        result += separator;
      }
    }
```

IvanFesenko (changes requested):

IvanFesenko (approved):

PR #2312, opened 2023-04-18
Solution

4ykis:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+
+    for (const item of this) {
+      if (item !== null && item !== undefined) {
+        res += item;
+      }
+
+      res += separator;
+    }
+
+    if (separatorInner || separatorInner.length) {
```
vlad-demchuk, line 19: this condition is always `true`, since you have a default value for `separator `

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+      res = res.slice(0, -separatorInner.length);
```
vlad-demchuk, line 20: can we avoid this `slice`
you can make this inside the loop

vlad-demchuk (changes requested):

src/arrayMethodJoin.js
VitaliyBondarenko1982: ok, in your solution we can't avoid if and slice. failed test with empty separator
also you can instead for of loop make with for loop and check current index. if it is last element - do not add a separator

VitaliyBondarenko1982: or another solution
remove if and return with slice
```
 return res
      .slice(0, -separatorInner.length || this.length);

VitaliyBondarenko1982 (approved):

PR #2315, opened 2023-04-18
solution

Absolution26:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const copyArr = [...this];
+    let arrJoin = '';
+    let sep = separator;
+
+    if (sep === undefined) {
+      sep = ',';
```
vlad-demchuk: `sep` can't be `undefined`. you set a default value for the argument

vlad-demchuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const copyArr = [...this];
```
polosanya: There is no need to make a copy, as you will not modify the array anyway

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+    let arrJoin = '';
+    let sep = separator;
+
+    if (sep === null) {
+      sep = 'null';
+    }
```
polosanya: This logic is redundant. It will work the same by default

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+
+    for (let i = 0; i < copyArr.length; i++) {
+      if (copyArr[i] === null || copyArr[i] === undefined) {
+        copyArr[i] = '';
+      }
```
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        arrJoin += this[i];
      }
```
Just do nothing if element is null or undefined

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+
+      if (i === copyArr.length - 1) {
+        arrJoin += copyArr[i];
+      } else {
+        arrJoin += copyArr[i] + sep;
+      }
```
polosanya: Same here. Just add separator only if it's not last element

polosanya (changes requested):
Almost perfect!

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,22 @@
...
+      if (i !== this.length - 1) {
+        arrJoin += separator;
+      }
+    }
+
+    return arrJoin;
   };
-}
+};
+// }
```
agniya-i, line 23: don't forget to remove comments

agniya-i (approved):

PR #2316, opened 2023-04-19
join

Dyakovich-R:


src/arrayMethodJoin.js
```diff
@@ -1,11 +1,25 @@
...
+        result += '';
+      } else {
+        result += this[i];
+      }
+
+      if (i < this.length - 1) {
+        result += separator === undefined ? ','
+          : separator === null ? 'null'
+            : typeof separator === 'number' && isNaN(separator) ? ','
+              : separator;
```
DarkMistyRoom: set default value to the `separator `instead of this complicated constraction

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,25 @@
...
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
DarkMistyRoom:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        result += this[i];
     }
```

DarkMistyRoom (changes requested):
let's improve the code

anstsot (approved):
Great job, approved!

PR #2317, opened 2023-04-19
Solution

Artem193:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,42 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const arrCopy = [...this];
```
IvanFesenko: No need to create copy

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,42 @@
...
+        string = string + arrCopy[i];
+      }
+
+      return string;
+    }
+
+    for (let i = 0; i < arrCopy.length - 1; i++) {
+      string = string + arrCopy[i] + ',';
+    }
+    string = string + arrCopy[arrCopy.length - 1];
```
IvanFesenko: No need to do so many operations. Algoritm should be next
create for loop, inside there are 2 if statement
one `if` statement check that this[i] not equal null or undefined if so - add to result `string += this[i]`
second `if` check if i is not last and add separator

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,42 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
```
IvanFesenko: Add default value for separator `function(separator = ',')`

IvanFesenko (changes requested):
Refactor required. Check my comments

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] != null) {
```
sTorba24: we should use strict equality here
```suggestion
      if (this[i] !== null) {
```

sTorba24: And what if value is undefined?

sTorba24 (changes requested):
GJ 🚀 Little fix needs to be done :)

qoa11a (approved):

PR #2320, opened 2023-04-21
 js_array-method-join

Donovanoff:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
anstsot: it can be simplified to
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        result += this[i];
      }
```

anstsot (changes requested):

anstsot (approved):
Great job, approved!

tetiana-ivanchenko (approved):

PR #2328, opened 2023-04-24
solution

RatiBorRozGa:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+
+    for (let i = 0; i < newArray.length; i++) {
+      result += newArray[i];
+
+      if (i !== newArray.length - 1) {
+        result += separator;
+      }
+    }
+
+    return result;
```
SanyaBratashchuk: You can simplify the solution. All you need to do is iterate over the array, checking at each iteration that the current element is neither null nor undefined and add it to the resulting string. also check that the current element is not the last element, if not, add a separator

SanyaBratashchuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
```suggestion
      if (this[i] !== null || this[i] !== undefined) {
        result += this[i];
      }
```

tetiana-ivanchenko (changes requested):

src/arrayMethodJoin.js
RatiBorRozGa: in this case i will take error in tests "Should work when array has null and undefined", because i will not replace null and undefined with empty element

DarkMistyRoom (approved):
Well done!

PR #2330, opened 2023-04-25
Solution of array method join

vanya-kalyenichenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+    let result = '';
+
+    if (this[0] !== null && this[0] !== undefined && this[0] !== '') {
+      result = `${this[0]}`;
+    }
+
+    if (this.length === 0) {
+      return '';
+    }
+
```
IvanFesenko: it's useless

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+    for (let i = 1; i <= this.length - 1; i++) {
+      result += separator;
+
+      if (this[i] === null || this[i] === undefined || this[i] === '') {
+        continue;
+      } else {
+        result += this[i];
+      }
+    }
```
IvanFesenko: You shoud add separator at the end in case when `i <= this.length - 1`
also you add to result only in case when `this[i] !== null && this[i] !== undefined`
```suggestion
    for (let i = 0; i < this.length; i++) {
      if (this[i] !== null && this[i] !== undefined) {
        // add to result 
      } 
      
      if (i < this.length - 1) {
        result += separator;
      }
    }
```

IvanFesenko (changes requested):
Check my comments

anstsot (approved):
Great job, approved!

PR #2333, opened 2023-04-25
Solution

anpiet:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,40 @@
...
+      }
+    } else if (this.length >= 2) {
+      for (let i = 1; i < this.length; i++) {
+        if (this[i] === null || this[i] === undefined) {
+          result += ',';
+        } else {
+          result += `${separator}${this[i]}`;
+        }
+      }
+    }
```
choeqq: you can also do that using 1 loop - [checklist ](https://github.com/mate-academy/js_array-method-join/blob/master/checklist.md) point 5
```suggestion
   for(let i = 0; i < this.length; i++) {
       if(this[i] !== null && this[i] !== undefined) {
          // add letter to our result
       }
       
       if(i is smaller than our length) {
         // add separator to new string
       }
   }
```

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,40 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
choeqq: please try to make it work with default separator set to `','`

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,40 @@
...
+        }
+      }
+    }
+
+    if (this[0] === null) {
+      return `${result}`;
+    } else {
+      return `${this[0]}${result}`;
+    };
   };
```
choeqq, line 29: please return result in the end

choeqq (changes requested):

mvjl000 (approved):

PR #2335, opened 2023-04-25
solution

HannieA:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
Anastasiia-Svintsova: Why do you need to add an empty string? Useless action
```suggestion
      if (this[i] !== null || this[i] !== undefined) {
        result += this[i];
      }
```

Anastasiia-Svintsova (changes requested):
Let's get rid of useless actions.

Anastasiia-Svintsova (approved):
👍

PR #2336, opened 2023-04-26
Join via for of

RostyslavKuts:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+    for (const mass of this) {
+      if (mass !== undefined && mass !== null) {
+        result += mass;
+      }
+
+      if (count !== this.length - 1) {
+        result += separator;
+      }
+
+      count++;
```
Anastasiia-Svintsova: It's better to use a simple `for` loop here. Then you'll have the index and the count will be useless.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    let count = 0;
+
+    for (const mass of this) {
```
Anastasiia-Svintsova: The `mass` name is not meaningful enough for the variable. Use more meaningful names, please

Anastasiia-Svintsova (changes requested):

polosanya (approved):
Well done!

PR #2338, opened 2023-04-27
MethodJoin

Dmytro-Tymchenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
```
OleziO: Remove all comments.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i !== 0) {
+        result += separator;
+      }
+
+      if (this[i] !== null && this[i] !== undefined) {
```
OleziO: Replace `this[i] !== null && this[i] !== undefined` with `this[i]`, it will be the same.

OleziO (changes requested):

src/arrayMethodJoin.js
Dmytro-Tymchenko: > Replace `this[i] !== null && this[i] !== undefined` with `this[i]`, it will be the same.

like this "if (this[i])" its not working

VitaliyBondarenko1982 (approved):

PR #2339, opened 2023-04-27
Solution

generatedman:


DidenkoDenis306 (changes requested):
Good job, but it's worth setting the default value "," to the separator, because function will not work if the separator is undefined

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,21 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
dinaminahi: @DidenkoDenis306 is right 👍 

this is how it works, when no separator is coming 
image

dinaminahi (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joined = '';
```
Anna-Tkachenko, line 8: Let's create a more descriptive variable name. For now, it's not clear what should be stored here. Probably `joinedValues` will be better

Anna-Tkachenko (approved):

PR #2341, opened 2023-04-28
Add solution

ghost:


generatedman (commented):
good job!

generatedman (approved):
good job!

naziamoff (changes requested):
good 👍 
But let's try a little different approach which improves the readability of your code:

Iterate over the whole array and handle it inside a loop, without the checks for the last element

naziamoff (approved):
🔥

PR #2344, opened 2023-04-30
finished cersion

EvhenIvchenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+
+      if (i === lastIndex) {
+        separatorChecked = '';
+      }
+
+      if (value === null || value === undefined) {
+        value = '';
+      }
+      string = string + value + separatorChecked;
+    }
```
```suggestion
    let string = '';
    let lastIndex = this.length - 1;

    for (let i = 0; i < this.length; i++) {
      if (this[i] !== null && this[i] !== undefined) {
        string += this[i];
      }
      
      if (i !== lastIndex) {
        string += separator;
      }
    }
```

tetiana-ivanchenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+    const lastIndex = this.length - 1;
```
Anastasiia-Svintsova, line 9: Don't really need to move this value to a variable. You use it only once

Anastasiia-Svintsova (approved):

src/arrayMethodJoin.js
EvhenIvchenko: Okay , thanks for your feedback

PR #2345, opened 2023-05-02
Solution

yuliia-nudyk:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,35 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
tetiana-ivanchenko: use dafault parametr value

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,35 @@
...
+
+    function typeCheck(char) {
+      if (char === undefined || char === null) {
+        return '';
+      } else {
+        return `${char}`;
+      }
+    }
+
+    return result;
```
tetiana-ivanchenko: let's simplify your solution, you need only 2 checks:
1. if  this[i] is not undefined and isn't null, add it to the result string
2. if it isn't last element, add separator
thats it

tetiana-ivanchenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 1) {
+      return `${this[0]}`;
+    }
```
polosanya: This check is actually redundant, as it will work the same with the loop anyway

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        result += `${this[i]}`;
```
```suggestion
        result += this[i];
```
There is no need to use template literals. The value will be converted into string anyway

polosanya (changes requested):
Almost perfect!

polosanya (approved):
Well done!

PR #2346, opened 2023-05-03
Solution

Pa1eOrc:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,24 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
     // write code here
+    let result = '';
+    let separatorCopy = separator;
+
+    if (separatorCopy === undefined) {
+      separatorCopy = ',';
+    }
```
anstsot: add the default value for `separator`

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,24 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
     // write code here
+    let result = '';
+    let separatorCopy = separator;
```
anstsot: do not create additional variable. just use `separator`

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,24 @@
...
+
+    if (separatorCopy === undefined) {
+      separatorCopy = ',';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        result += separatorCopy;
+      }
```
anstsot: also do not add the separator after the last element

anstsot (changes requested):

SanyaBratashchuk (approved):

PR #2348, opened 2023-05-05
Solution

1v43k:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    let separ = separator;
```
anstsot: do no reassign `separ` jut use the `separator`

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
...
+
+    if (separator === undefined) {
+      separ = ',';
+    }
```
anstsot: set the default value for the argument

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        result += separ;
+      }
```
anstsot: do not add separator after last value

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
...
+
+      if (this[i] !== undefined && this[i] !== null) {
+        result += this[i];
+      } else {
+        result += '';
+      }
```
anstsot: can be simplified
```suggestion
      if (this[i] !== undefined && this[i] !== null) {
        result += this[i];
      }
```

anstsot (changes requested):

etojeDenys (approved):

PR #2349, opened 2023-05-05
ArrayMethod_Join_complete

ZupaFly:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
+        result += ',';
+      } else if (i > 0 && separator !== undefined) {
+        result += separator;
+      }
+
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
witflash: You can do it simpler:
```
  result += this[i];
}
```
And you can avoid using `else` block.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0 && separator === undefined) {
+        result += ',';
+      } else if (i > 0 && separator !== undefined) {
+        result += separator;
+      }
```
witflash: You can set a default value for your `separator` argument (`function(separator = ',')`). After this, you can check if the current iteration is not last and add this separator after `if` block below.

witflash (changes requested):
Good job!
But let's try to improve your code a little bit.

ivanchuk-yurii (approved):

PR #2352, opened 2023-05-06
add task solution

maksym-nemera:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+        newSeparator = '';
+      }
+
+      if (this[i] === null) {
+        arrayCopy[i] = '';
+      }
+
+      if (this[i] === undefined) {
+        arrayCopy[i] = '';
+      }
```
DidenkoDenis306: t would be better to combine this into one if statement using the && operator

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const arrayCopy = [...this];
```
DidenkoDenis306, line 8: you don't need to create a copy of the array

DidenkoDenis306 (changes requested):
Good job))) 
But there are a couple of things that need to be fixed.

src/arrayMethodJoin.js
maksym-nemera: @DidenkoDenis306  If I don't create copy I'll change the origin input

src/arrayMethodJoin.js
maksym-nemera: @DidenkoDenis306 

yeah for lines 19 - 25, it can be, but with the || operator. I'll do it

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+    const len = this.length;
+
+    let newSeparator = '' + separator;
```
```suggestion
    let newSeparator = String(separator);
```

silchyn (approved):

PR #2353, opened 2023-05-07
Solution

VolodymyrPst:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,41 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
```
OleziO, line 8: Remove all comments.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,41 @@
...
+    const arr = [...this];
+    const chengeddArr = [];
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        arr[i] = '';
+      }
+    }
+
+    if (this.length === 0
```
OleziO, line 19: Replace `this.length === 0 || this === undefined || this === null` with `!this.length`

OleziO (changes requested):

PR #2354, opened 2023-05-07
it works

zisuan-od:
I don't like it, but it works
how can I make it better?


vlad-demchuk (changes requested):
Replied in fe_chat about how to improve

mykhalenych (approved):

PR #2356, opened 2023-05-08
Develop

VolodymyrPst:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,40 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
     // write code here
```
OleziO, line 8: Remove all comments from the code.

OleziO (changes requested):

PR #2357, opened 2023-05-08
Develop

VolodymyrPst:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,40 @@
...
+      return '';
+    }
+
+    for (const item of arr) {
+      if (separator === null) {
+        chengeddArr.push(item);
+        chengeddArr.push('null');
+      } else if (separator === undefined) {
+        chengeddArr.push(item);
+        chengeddArr.push(',');
```
nazarmatsevych: You can set comma as a default separator

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,40 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,40 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        arr[i] = '';
+      }
+    }
+
+    if (!this.length) {
+      return '';
+    }
```
nazarmatsevych, line 20: Looks too complicated, actually we only need two if's to check current element for null and undefined and check current index

```
        str += this[i];
      }

      if (i !== this.length - 1) {
        str += separator;
      }
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,40 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let str = '';
+    const arr = [...this];
+    const chengeddArr = [];
```
nazarmatsevych, line 10: redundant

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,40 @@
...
+        chengeddArr.push(item);
+        chengeddArr.push(separator);
+      }
+    }
+
+    chengeddArr.length = chengeddArr.length - 1;
+
+    for (const item of chengeddArr) {
+      str += item;
+    }
```
nazarmatsevych: Redundant loops, we can combine this checks inside one loop

nazarmatsevych (changes requested):

PR #2358, opened 2023-05-08
Develop

VolodymyrPst:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
...
+
+    for (const item of this) {
+      if (separator === null) {
+        chengeddArr.push(item);
+        chengeddArr.push('null');
+      } else {
+        chengeddArr.push(item);
+        chengeddArr.push(separator);
+      }
+    }
```
nazarmatsevych: What for we need this second loop?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+    const chengeddArr = [];
```
nazarmatsevych: What for we need this array?

nazarmatsevych (changes requested):

PR #2361, opened 2023-05-08
crate join

pos1bl:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    let separ = 'null';
```
volodymyr-soltys97: Why do you create a new const separ? If you have separator as props. Try to use only separator in props

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      let value = this[i];
+
+      if (value === undefined || value === null) {
+        value = '';
+      }
+
+      result += value;
```
volodymyr-soltys97: Here it is better to check for not equal and immediately add the result inside. 
Try to rewrite this

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more readable

Viktor-Kost (approved):

PR #2364, opened 2023-05-09
Solution

snuber-w:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,43 @@
...
+        newSeparator = ',';
+        break;
+
+      case (null):
+        newSeparator = 'null';
+        break;
+
+      default:
+        newSeparator = separator.toString();
+    }
```
DidenkoDenis306: It would be better to rewrite this to if statement

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,43 @@
...
+    if (this[0] === null || this[0] === undefined) {
+      resultString = '';
+    } else {
+      resultString = this[0].toString();
+    }
+
+    switch (newSeparator) {
+      case (undefined):
+        newSeparator = ',';
+        break;
```
DidenkoDenis306: You can give default value ',' to separator instead

DidenkoDenis306 (changes requested):
Quite an unusual implementation of this method and it works, it's cool! But it would be nice to change some things.

src/arrayMethodJoin.js
snuber-w: that makes sense, thanks

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,37 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
```
dinaminahi: No need for this `if`, see Checklist p.5 - Remember that loop with false condition simply will iterate 0 times.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,37 @@
...
+
+    let resultString = '';
+    let newSeparator = separator;
+
+    if (this[0] === null || this[0] === undefined) {
+      resultString = '';
+    } else {
+      resultString = this[0].toString();
+    }
```
dinaminahi: No need for two different checks for first element and the rest, you can start your loop with 0 and check first element same way as other elements

dinaminahi (changes requested):

src/arrayMethodJoin.js
snuber-w: 5. [CODE STYLE] - don't write loops inside if statement. It's making your code too complicated. Remember that loop with false condition simply will iterate 0 times.
it says not to use a loop inside if statement. Am I misunderstanding something?

33Orange (approved):
Good job! ⭐

PR #2366, opened 2023-05-10
Solution

BogdanParubchenk01:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+      return '';
+    }
+
+    let result = `${this[0]}`;
+
+    if (this[0] === null) {
+      result = '';
+    }
+
+    if (this.length === 1) {
```
Viktor-Kost: This condition is also not needed because the loop in line 24 will not execute if 'this.length' is 1.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    // if (typeof separator !== 'string') {
```
Viktor-Kost: The commented code should be removed if it's not needed.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+    //   return this;
+    // }
+    if (this.length === 0) {
+      return '';
+    }
+
+    let result = `${this[0]}`;
+
+    if (this[0] === null) {
```
Viktor-Kost: This condition is not needed because you have already covered the case when 'this[0]' is null in the 'result' initialization.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    // if (typeof separator !== 'string') {
+    //   return this;
+    // }
+    if (this.length === 0) {
+      return '';
+    }
+
+    let result = `${this[0]}`;
```
Viktor-Kost: Consider using a ternary operator to initialize 'result' with the appropriate value in one step: let result = (this[0] === null) ? '' : `${this[0]}`;

Viktor-Kost (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
+
+    let result = (this[0] === null) ? '' : `${this[0]}`;
+
+    for (let i = 1; i < this.length; i++) {
```
DarkMistyRoom, line 14:
```suggestion
    let result = '';

    for (let i = 0; i < this.length; i++) {
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+      if (this[i] === null || this[i] === undefined) {
+        result += `${separator}`;
+      } else {
+        result += `${separator}` + this[i];
+      }
```
DarkMistyRoom, line 19:
```suggestion
      if (this[i] !== null || this[i] !== undefined) {
        result += this[i]`;
      }
      
      if (i + 1 !== this.length) {
        result  += separator;
      }
```

src/arrayMethodJoin.js
DarkMistyRoom: to reduce the code

src/arrayMethodJoin.js
DarkMistyRoom: to avoid duplicated adding of the `separator`, and `if else` nesting

DarkMistyRoom (approved):

PR #2368, opened 2023-05-10
solution

skinnyjazz:
"I just can't understand why both options are possible in this task on line 12?"
`if (this[i] !== null && this[i] !== undefined) {`
`if (this[i] != null) {`

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const lastIndex = this.length - 1;
+    let newString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        newString += `${this[i]}`;
+      }
+
+      if (i !== lastIndex) {
+        newString += `${separator}`;
+      }
+    }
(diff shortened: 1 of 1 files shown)


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const lastIndex = this.length - 1;
+    let newString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
```
cephy77, line 12: if (this[i] != null) is works because null === undefined. And when you have a non-strict comparison it seeking not only null

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+        newString += `${this[i]}`;
```
cephy77, line 13: It works, but interpolation is not necessary here

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+      }
+
+      if (i !== lastIndex) {
+        newString += `${separator}`;
```
cephy77, line 17: And the separator also does not need the interpolation

cephy77 (approved):
That's the way to do it 😎 !)

silchyn (approved):

PR #2369, opened 2023-05-10
add join method

AleksandrTsivilin:


src/arrayMethodJoin.js
```diff
@@ -6,7 +6,47 @@
...
+    return separator.toString();
+  }
+
+  if (separator === '') {
+    return separator;
+  }
+
+  if (separator === null) {
+    return 'null';
+  }
```
dinaminahi: No need for such a big check, because if you add something to string it become a string 
image
image

src/arrayMethodJoin.js
```diff
@@ -6,7 +6,47 @@
...
+
+  return ',';
```
dinaminahi: Instead of this, you can set default value for a separator like this: 

`
[].__proto__.join2 = function(separator = ',' ) {
`

src/arrayMethodJoin.js
```diff
@@ -6,7 +6,47 @@
...
   [].__proto__.join2 = function(separator) {
     // write code here
+    const delimiter = getDelimiter(separator);
+
+    let output = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const current = isValidItem(this[i])
+        ? this[i]
+        : '';
```
dinaminahi: Can be simplified like output += isValidItem(this[i]) ? this[i] : '';

src/arrayMethodJoin.js
```diff
@@ -6,7 +6,47 @@
...
+      const part = !isLastIndex(this, i)
+        ? current + delimiter
+        : current;
+
+      output += part;
```
dinaminahi: and same here

dinaminahi (changes requested):

src/arrayMethodJoin.js
AleksandrTsivilin: I guess It is more readable

AleksandrTsivilin: Anyway I will fix it

src/arrayMethodJoin.js
AleksandrTsivilin: I got a mistake of the linter ( length > 80 characters) What do you recommend me

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,31 @@
...
+    }
+
+    return output;
   };
 }
 
+function _isValidItem(item) {
```
```suggestion
function isItemValid(item) {
```

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,31 @@
...
+  return item !== undefined && item !== null;
+}
+
+function _isLastIndex(array, index) {
```
```suggestion
function isLastIndex(array, index) {
```

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,31 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    const delimiter = separator;
```
silchyn: is it necessary?

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,31 @@
...
+
+    let output = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const current = _isValidItem(this[i]) ? this[i] : '';
+      const part = !_isLastIndex(this, i)
+        ? current.toString() + delimiter
+        : current;
+
+      output += part;
```
```suggestion
      if (isItemValid(this[i])) {
        output += this[i].toString();
      }
      
      if (!isLastIndex(this, i)) {
        output += delimiter;
      }
```

silchyn (changes requested):
Sometimes variables can be excessive

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,32 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    // const delimiter = separator;
```
naziamoff: never leave comments in your code unless you really need to

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,32 @@
...
   };
 }
 
+function isItemValid(item) {
+  return item !== undefined && item !== null;
+}
+
+function isLastIndex(array, index) {
+  return index === array.length - 1;
+}
```
naziamoff: this one is definitely redundant here

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,32 @@
...
+      if (!isLastIndex(this, i)) {
+        output += separator;
+      }
+    }
+
+    return output;
   };
 }
 
+function isItemValid(item) {
```
naziamoff: and this one looks more like a variable but not a function, I believe

naziamoff (changes requested):

naziamoff (approved):

PR #2370, opened 2023-05-11
add functionality to applyCustomJoin

cephy77:


ghost (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let element = '';
```
33Orange: You don't need this variable

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+
+      if (this[i] != null) {
```
33Orange: Update condition to check if item not equal to `null` and `undefined`.  
with data type checking (`!==`)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+        element = this[i];
```
33Orange: You don't need to overwrite new variable `element` here. You can overwrite `string` directly

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+      }
+
+      string += element;
```
33Orange: This will not be used if you rewrite `string` inside an `if` condition

33Orange (changes requested):
Good job.

A few moments to fix.

silchyn (approved):

PR #2371, opened 2023-05-11
solution

artem-ivashchenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let copyElement = this[i];
+
+      if (this[i] == null) {
```
Anastasiia-Svintsova: Use strict equality '===' instead of '==' for comparison. And what will be in case this[i] will be undefined?

Anastasiia-Svintsova (changes requested):
Good job! Let's add some improvements

src/arrayMethodJoin.js
artem-ivashchenko: when == used , undefined == null, but undefined !== null. I just didn't want to write the long condition if( this[i] === undefined || this[i] === null), should i use that anyway?

sTorba24 (approved):
Niiice!

PR #2372, opened 2023-05-11
solution

okosohor:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
+      switch (word) {
+        case undefined:
+        case null:
+          break;
+
+        default:
+          joinResult += word;
+      }
+
+      if (word !== this[this.length - 1]) {
```
33Orange: This condition may cause incorrect results for repeated elements. Instead, use the loop index to check if it's the last iteration.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
+
+    for (const word of this) {
+      switch (word) {
+        case undefined:
+        case null:
+          break;
+
+        default:
+          joinResult += word;
+      }
```
33Orange: Use 1 `if` statement here. It would be more understandable.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let joinResult = '';
+    let newSeparator = separator;
+
+    if (newSeparator === undefined) {
+      newSeparator = ',';
+    }
```
33Orange: You can give default value ',' to separator instead

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
+
+    for (const word of this) {
```
33Orange: Use default `for` loop here in order to have easy access to the index.

33Orange (changes requested):
Good job.

Some comments:
- You can give default value ',' to separator.
- Use default `for` loop here in order to have easy access to the index. And this update will change `if` condition on line 25.
- Remove `switch` statement.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinResult = '';
+
+    // for (const word of this) {
+    //   if (word !== undefined && word !== null) {
+    //     joinResult += word;
+    //   }
```
dinaminahi, line 13: Redundant comment

dinaminahi (approved):
Great job 👍 
Don`t forget to delete redundant comments

PR #2373, opened 2023-05-11
Solution

vlkzmn:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,34 @@
...
+    switch (true) {
+      case (this[0] === null || this[0] === undefined || !this):
+        result = '';
+        break;
+      case (typeof (this[0]) === 'number'):
+        result = this[0].toString();
+        break;
+      default:
+        result = this[0];
+    }
```
anstsot: this code seems to be redundant. why do you need it?

anstsot (changes requested):

witflash (approved):
Well done 👍

PR #2375, opened 2023-05-11
add solution

niaYaro:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,25 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const joined = [...this];
```
Anna-Tkachenko: It's not something joined. In this variable you've stored the copy of original array

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,25 @@
...
+      joinedSring += String(joined[i]) + separator;
+    }
+
+    switch (separator) {
+      case '':
+        return joinedSring.slice(0, joinedSring.length);
+
+      default:
+        return joinedSring.slice(0, joinedSring.lastIndexOf(separator));
+    }
```
Anna-Tkachenko: Instead of using additional `switch / case` here, you can simply add some check inside the loop for checking should you add separator or no

Anna-Tkachenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedStr = '';
+    const copyOfOriginal = [...this];
```
dinaminahi, line 9: You could use original array, because you are not mutating it here

dinaminahi (approved):

PR #2376, opened 2023-05-11
solution

Vepr984:


src/arrayMethodJoin.js
```diff
@@ -3,9 +3,29 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else if (typeof this[i] === 'string'
+      && this[i] === 'null') {
+        result += 'null';
+      } else if (isNaN(this[i])) {
```
etojeDenys: This condition is unnecessary. If the value is not a string, it will be implicitly converted to a string when concatenated to 'result'.

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,29 @@
...
+        result += 'NaN';
+      } else {
+        result += this[i];
+      }
```
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        result += this[i];
      } 
```

etojeDenys (changes requested):

YegorKorenkov (approved):
Nice work!

PR #2381, opened 2023-05-12
Solution

dimoshen:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,43 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    const length = this.length;
+    let firstElement = String(this[0]);
+    let separatorValue = separator;
+
+    if (length === 0) {
+      return '';
+    }
+
+    if (firstElement === 'undefined' || firstElement === 'null') {
```
dinaminahi: First element shouldn't have separate check and can be checked along with others

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,43 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const length = this.length;
+    let firstElement = String(this[0]);
+    let separatorValue = separator;
+
+    if (length === 0) {
```
dinaminahi: No need for this check, if the length is 0, the for loop won`t start

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,43 @@
...
+
+    if (firstElement === 'undefined' || firstElement === 'null') {
+      firstElement = '';
+    }
+
+    if (length === 1) {
+      return firstElement;
+    }
+
+    if (separatorValue === undefined) {
```
dinaminahi: There is no need to check if it is undefined because you gave it default value, which means that if it is undefined it gets value of default value

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,43 @@
...
+      separatorValue = ',';
+    }
+
+    if (separatorValue === null) {
+      separatorValue = 'null';
+    }
+
+    let result = firstElement;
```
dinaminahi: So let's set initial value to empty string

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,43 @@
...
+
+    if (length === 1) {
+      return firstElement;
+    }
+
+    if (separatorValue === undefined) {
+      separatorValue = ',';
+    }
+
+    if (separatorValue === null) {
```
dinaminahi: Also no need for this check, because if separator is null and you add it to your joined string, it will become a string
image

dinaminahi (changes requested):

src/arrayMethodJoin.js
dimoshen: fixed

src/arrayMethodJoin.js
dimoshen: right. got rid of it

src/arrayMethodJoin.js
dimoshen: removed

src/arrayMethodJoin.js
dimoshen: fixed

src/arrayMethodJoin.js
dimoshen: got it

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const length = this.length;
```
naziamoff: no need for this variable

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+    const separatorValue = separator;
```
naziamoff: and this one as well

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+    for (let i = 0; i < length; i++) {
+      let element = this[i];
+
+      if (element === undefined || element === null) {
+        element = '';
+      }
+
+      if (i === length - 1) {
+        result += element;
+        break;
```
naziamoff: no need for a break here. Basically the loop will break itself, as it is the last iteration

naziamoff (changes requested):
not bad, but let's try more clear logic:

- inside the loop add the element only if it is not null or undefined
- if the iteration is not the last one - add separator

src/arrayMethodJoin.js
dimoshen: removed

src/arrayMethodJoin.js
dimoshen: removed

src/arrayMethodJoin.js
dimoshen: removed any breaks

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const element = this[i];
+
+      if (element !== undefined && element !== null) {
+        result += element;
+      } else {
+        result += '';
+      }
```
33Orange: `const element = this[i];` redunant here.

Inside your condition `(element !== undefined && element !== null)` you can directly add a string to the result if it passes this condition.

This will remove the extra `else` statement that does nothing
```
else {
  result += '';
}
```

33Orange (changes requested):
Almost done!

Just one moment with readability !

src/arrayMethodJoin.js
dimoshen: fixed

dinaminahi (approved):
Well done 👍

naziamoff (approved):

naziamoff (approved):

naziamoff (approved):

PR #2383, opened 2023-05-12
arrayMethodJoin solution

panch1811:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (separator !== undefined && i !== 0) {
```
dinaminahi, line 12: There is no need to check if it is `undefined` because you gave it default value, which means that if it is `undefined` it gets value of default value

dinaminahi (approved):

naziamoff (approved):

PR #2385, opened 2023-05-12
Solution

vchervinko:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joined = '';
```
Anna-Tkachenko: Let's create mode descriptive variable name
For now it's not clear joined `what` should be stored here 🐝

Anna-Tkachenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
```
33Orange, line 7: It's better to avoid modifying the native Array prototypes directly. Instead, create a new function and extend the Array.prototype with it.

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,21 @@
...
+    let joinedString = '';
+
+    for (let i = 0; i < this.length; i++) {
```
33Orange, line 10: Consider using 'for...of' loop to make the code more readable by getting rid of index access.

33Orange (approved):
Well done ! 🔥

PR #2388, opened 2023-05-12
Solution

ivankhrsh:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
```
33Orange, line 7: Consider using Object.setPrototypeOf() instead of modifying the Array prototype directly. This will help to avoid conflicts with other code.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+    let resultString = '';
+
+    for (let i = 0; i < this.length; i++) {
```
33Orange, line 10: Use 'let' instead of 'const' for the loop index variable 'i', because its value will change during the loop.

33Orange (approved):
Well done ! 🔥

PR #2389, opened 2023-05-12
Solution

bohdan-mykhailenko:


src/arrayMethodJoin.js
```diff
@@ -3,9 +3,26 @@
 /**
  * Implement method join
  */
+
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+    const lastIndexOfElement = this.length - 1;
```
```suggestion
    const lastIndex = this.length - 1;
```

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,26 @@
...
+    const emptyValues = [null, undefined];
```
```suggestion
    const ignoredValues = [null, undefined];
```

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,26 @@
...
+
+    for (const element of this) {
```
```suggestion
    for (const character of this) {
```

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,26 @@
...
+      const indexOfElement = this.indexOf(element);
+
+      if (!emptyValues.includes(element)) {
+        joinedString += element;
+      }
+
+      if (indexOfElement !== lastIndexOfElement) {
```
silchyn: and what if the initial string has duplicates?

silchyn (changes requested):

naziamoff (approved):
awesome 🔥

silchyn (approved):

PR #2391, opened 2023-05-12
add arrayMethodJoin task solution

booobryyyk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,36 @@
...
+
+    if (separator === undefined) {
+      separator = ',';
+    }
+
+    if (separator === null) {
+      separator = 'null';
+    }
+
+    for (let i = 0; i < newArr.length; i++) {
```
33Orange: Instead of modifying the original array, directly add the processed element to the result string.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,36 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(initialSeparator) {
+    let joinedArrString = '';
+    let separator = initialSeparator;
+    const newArr = [...this];
```
33Orange: Avoid creating a new array using the spread operator. You can directly use the original array.
But avoid array mutation

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,36 @@
...
+
+    if (separator === undefined) {
+      separator = ',';
+    }
+
+    if (separator === null) {
+      separator = 'null';
+    }
```
33Orange: You can get rid of this by simply adding a default value for the parameter.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,36 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(initialSeparator) {
```
33Orange: Add default value for `initialSeparator` (`initialSeparator = ','`)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,36 @@
...
+    for (let i = 0; i < newArr.length; i++) {
+      if (newArr[i] === null) {
+        newArr[i] = '';
+      }
+
+      if (newArr[i] === undefined) {
+        newArr[i] = '';
+      }
+
+      joinedArrString += newArr[i];
```
33Orange: You can combine all this logic into one if condition.
Just use `!==` to determine if there is a value and if it is - add it to the string

33Orange (changes requested):
Good job.

A few moment to fix.

naziamoff (approved):
🔥

PR #2392, opened 2023-05-12
array_join

DenisGurskiy:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
  * Implement method join
```
OleziO, line 4: Remove all comments from the code.

OleziO (approved):

PR #2393, opened 2023-05-12
[ym] join

YeMyr:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length < 1) {
+      return '';
+    }
+
+    if (this.length < 2) {
+      return String(this[0]);
+    }
```
silchyn: is it necessary?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+
+    const gap = (separator === undefined || separator === null)
+      ? `${separator}`
+      : separator;
```
silchyn: same, is this check necessary?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+
+    let output = (Object.is(this[0], null) || Object.is(this[0], undefined))
+      ? ''
+      : this[0];
```
silchyn: same here

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+
+    for (let i = 1; i < this.length; i++) {
+      const item = (Object.is(this[i], null) || Object.is(this[i], undefined))
+        ? ''
+        : this[i];
```
silchyn: and also here

silchyn (changes requested):
Think about generally cleaner code

src/arrayMethodJoin.js
YeMyr: yes, this check is needed, otherwise the function writes null and undefined as a string item ```'null' & 'undefined'```, instead of ```'<empty value>'``` 
So, I made it more readable and put inside the loop
fixed 👌

src/arrayMethodJoin.js
YeMyr: fixed 👌
removed

src/arrayMethodJoin.js
YeMyr: fixed 👌
removed

src/arrayMethodJoin.js
YeMyr: fixed 👌
removed at all

33Orange (changes requested):
Good job !

Let's try to improve readability a bit.
Let's check if the value is not `null` and `undefined` and immediately add it to the result string.
And after that, make 1 check to see if it the last element of the array or not and add a `separator` if not

YeMyr:
@33Orange 
Hi! thanks for comment, I've made changes, please, check it

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let output = '';
+
+    for (let i = 0; i < this.length; i++) {
+      output += this[i] === null || this[i] === undefined
+        ? ''
+        : this[i];
```
naziamoff: ternary is not necessary here. You can just add the element if `this[i] !== null or undefined`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+
+      if (i !== this.length - 1) {
+        output += `${separator}`;
```
naziamoff: what happens if we add some value to the string? You should check this yourself

naziamoff (changes requested):
not bad, but let's make some extra fixes

src/arrayMethodJoin.js
YeMyr: this is an artifact from previous versions in which null did not want to turn into a string 🤷‍♂️
thanks! fixed 👌

src/arrayMethodJoin.js
YeMyr: thanks for the notice 
fixed 👌

src/arrayMethodJoin.js
naziamoff: hehe, got it 😄

naziamoff (approved):
awesome 🔥

silchyn (approved):

PR #2395, opened 2023-05-13
Solution

k1samehoshigaki:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let updatedData = '';
```
silchyn: which data is updated?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+    const separatorToStr = String(separator);
```
silchyn: looks like function name, not variable

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+    let arrayValue;
```
silchyn: does it have to be here and what `arrayValue` means?

silchyn (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let updatedArrayData = '';
+    // let arrayValue;
```
33Orange, line 9: Remove the commented code (let arrayValue;) as it is not being used.

33Orange (approved):
Well done!
**Don't forget to clean up the code from unnecessary comments and everything that doesn't work in the code before pushing it.**

silchyn (approved):

PR #2399, opened 2023-05-14
Solution

julshvets:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let item = this[i];
+
+      if (this[i] === undefined || this[i] === null) {
+        item = '';
+      }
```
33Orange: You do not need to declare the value once again and overwrite it if null or undefined comes to you.
You can use `!==` in `if` condition and if value exist add it to result string.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+
+      if (i === this.length - 1) {
+        resultString += `${item}`;
+      } else {
+        resultString += `${item}${separator}`;
+      }
```
33Orange: Just use `if` condition, without `else`. If you take the previous step, you can only use `if` condition here

33Orange (changes requested):
Good job.

A few moments to fix.

src/arrayMethodJoin.js
julshvets: According to the description of the join method, if the element is undefined or null, it must be converted to an empty string instead of "null" or "undefined". This if condition converts the following values ​​to an empty string. Then we have to add this empty string to the result string. Otherwise, the solution does not pass tests.

33Orange (changes requested):
You didn't fix code according to my prev comment

src/arrayMethodJoin.js
33Orange: This code redunant here. Just check what you get when you use built-in `join` method, when you nave `null` or `undefined` in your array.

julshvets: Thank you!

dinaminahi (approved):
Well done 👍

PR #2400, opened 2023-05-14
Solution

SofiaLiubchenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringFromArray = '';
```
naziamoff: let's just stick to `string` here. The name you've selected overcomplicates it a little

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+
+    let workingSeparator = separator;
+
+    if (workingSeparator === null) {
+      workingSeparator = 'null';
+    }
```
naziamoff: do you really need this check?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      let element = this[i];
+
+      if (element === null || element === undefined) {
+        element = '';
+      }
```
naziamoff: let's try different approach - add the element to the resulting string only if it is not null or undefined

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+
+      if (i === this.length - 1) {
+        stringFromArray += element;
+      } else {
+        stringFromArray += element + workingSeparator;
+      }
```
naziamoff: same here - you can handle it inside one `if` if you check for the last index

naziamoff (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    let workingSeparator = separator;
+
+    if (workingSeparator === null) {
+      workingSeparator = 'null';
+    }
```
33Orange: Remove redundant check.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    let workingSeparator = separator;
```
33Orange: You don't need to copy separator.
Use `separator` from arguments.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+    if (workingSeparator === null) {
+      workingSeparator = 'null';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      let element = this[i];
+
+      if (element === null || element === undefined) {
+        element = '';
+      }
```
33Orange: Change your `if` statement.
Add `element` to `string` only if `element` !== `null` or !== `undefined` .
Check @naziamoff comments

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+
+      if (i !== this.length - 1) {
+        string += element + workingSeparator;
+      } else {
+        string += element;
+      }
```
33Orange: In this `if` statement, check if this is the last item in the array. If not, then add a separator to the string.
You only need to add a separator here. We will expand `string` above.

33Orange (changes requested):
Check @naziamoff comments.
You need to change your `if` conditions.

naziamoff (approved):
well done!

PR #2401, opened 2023-05-14
solution

kirilshorin:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const copyArray = [...this];
```
naziamoff: do you really need this copy? What for?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+
+    for (let i = 0; i < copyArray.length; i++) {
+      if (copyArray[i] === null || copyArray[i] === undefined) {
+        copyArray[i] = '';
+      }
```
naziamoff: you're basically telling that if the char is null or undefined - `do nothing`
Let's try a different approach instead - add the char to the result if it is a valid string

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+
+      if (i === copyArray.length - 1) {
+        result = result + copyArray[i];
+      } else {
+        result = result + copyArray[i] + separator;
+      }
```
naziamoff: and with this approach here you can just add separator here instead of those long checks

naziamoff (changes requested):
not bad already, but let's polish a little 💅

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        (i === this.length - 1)
+          ? result = result + this[i]
+          : result = result + this[i] + separator;
+      } else {
+        (i === this.length - 1)
+          ? result += ''
+          : result += '' + separator;
```
naziamoff: do you really need to add an empty line to the result if the current element is the last one?

naziamoff (changes requested):
it is possible to do with only to `ifs`, without `else`. Let's find out how

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        (i === this.length - 1)
+          ? result = result + this[i]
+          : result = result + this[i] + separator;
+      } else if (i !== this.length - 1) {
+        result += '' + separator;
+      }
```
silchyn, line 17:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        result += this[i];
      }
      
      if (i !== this.length - 1) {
        result += String(separator);
      }
```

silchyn (changes requested):

naziamoff (approved):

PR #2404, opened 2023-05-14
solution

annholovko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const joinSeparator = separator;
```
naziamoff: does this variable serve some higher purpose? 😄 
If it does not - you don't need it

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+
+    const lastIndex = this.length - 1;
+
+    let joinedString = '';
+
+    if (this.length === 0) {
+      return '';
+    }
+
```
naziamoff: you don't need this check, because if there's no length - the loop just won't start iterating

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const joinSeparator = separator;
+
+    const lastIndex = this.length - 1;
+
+    let joinedString = '';
```
naziamoff: I believe it can be just string
```suggestion
    let string = '';
```

naziamoff (changes requested):
well done 🔥 
fix a couple of small things and we're good:

naziamoff (approved):
awesome 🔥

PR #2410, opened 2023-05-15
Solution added

olehmahera:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const arrLength = this.length;
```
naziamoff, line 9: I would say this variable is redundant

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
naziamoff, line 8: let's give more descriptive names to our variables - it should be maximally clear what is it about

naziamoff (approved):
looks good 👍 
But consider my comment about the naming in your next PRs

PR #2413, opened 2023-05-15
Solution

myplekan:


src/arrayMethodJoin.js
```diff
@@ -3,9 +3,42 @@
...
+    let str = '';
+    let sep = separator;
+
+    if (separator === null) {
+      sep = 'null';
+    }
+
+    if (sep === undefined) {
+      sep = ',';
+    }
```
vlad-demchuk: let's use a default value for function argument to avoid these checks

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,42 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1 && this[i] === undefined) {
+        str += '';
```
vlad-demchuk: why do we need ` str += '';`?

vlad-demchuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,36 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+    let sep = separator;
+
+    if (separator === null) {
+      sep = 'null';
+    }
```
Viktor-Kost: redundant since you use a default value
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,36 @@
...
+
+        return str;
+      }
+
+      if (this[i] === undefined || this[i] === null) {
+        str += ',';
+        continue;
+      }
+
+      str += this[i] + sep;
```
Viktor-Kost: you can write not so many conditions. Do like this 
```suggestion
      if (this[i] !== undefined && this[i] !== null) {
        str += this[i];
      }

      if (i < this.length - 1) {
        str += separator;
      }
```

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,36 @@
 /**
  * Implement method join
  */
+
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+    let sep = separator;
```
Viktor-Kost: redundant
```suggestion
```

Viktor-Kost (changes requested):

volodymyr-soltys97 (approved):
Great job 🔥

PR #2414, opened 2023-05-15
solution

Gosyav:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    let divider = separator;
+
+    if (divider === undefined) {
+      divider = ',';
```
Viktor-Kost: This condition is redundant, as the default value for 'separator' has already been set in the function definition. You can remove lines 11-15.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+      divider = String(separator);
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        result += divider;
+      }
+
+      if (this[i] !== null && this[i] !== undefined) {
+        result += String(this[i]);
```
Viktor-Kost:
```suggestion
        result += this[i];
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    let divider = separator;
```
Viktor-Kost: redundant variable. You can use separator directly at 19 line

Viktor-Kost (changes requested):

etojeDenys (approved):
awesome

PR #2416, opened 2023-05-16
Solution

mkorulczyk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function (separator = ',') {
+    let joinedArray = '';
+    const lastIndex = this.length - 1;
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] != null) {
```
DorotaLeniecDev, line 12: Consider also scenario when undefined is a value :)

DorotaLeniecDev (changes requested):
Good job, one more change needed

PR #2426, opened 2023-05-21
join-Is-Done

EvgenGolub:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const arr = [...this];
+
+    let str = '';
+
+    if (arr.length === 0) {
+      return '';
+    }
```
nazarmatsevych: Redundant check, you already have it inside your loop

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const arr = [...this];
```
nazarmatsevych: No need to make a copy, you can use this for iteration

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+    if (arr.length === 0) {
+      return '';
+    }
+
+    for (let i = 0; i < arr.length; i++) {
+      if (arr[i] === null || arr[i] === undefined) {
+        arr[i] = '';
+      }
+
+      str += arr[i];
```
```suggestion
      if (arr[i] !== null && arr[i] !== undefined) {
        str += arr[i];
      }
```

nazarmatsevych (changes requested):

vlad-demchuk (approved):

PR #2429, opened 2023-05-23
Solution

BeeGooDee:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+      if (arr[i] === undefined || arr[i] === null) {
+        arr[i] = '';
+      }
+    }
+
+    let str = `${arr[0]}`;
+
+    for (let i = 1; i < arr.length; i++) {
+      str += `${separator}${arr[i]}`;
+    }
```
sTorba24: you can minimize it to one loop. if add arr[i] or '', then separator. If last element - don't add separator

sTorba24 (changes requested):
GJ 🚀 Lets improve it

witflash (approved):
Good job! 👍

PR #2430, opened 2023-05-23
Solution

reseneweb:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,31 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```
Just set the default value for separator

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,31 @@
...
+    let result = '';
+    let delimiter = separator;
+
+    if (delimiter === undefined) {
+      delimiter = ',';
+    }
+
+    if (delimiter === null) {
+      delimiter = 'null';
+    };
```
```suggestion
```
Null will be converter into string 'null' anyway

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,31 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (((this[i] === null) || (this[i] === undefined))
+       && (i !== (this.length - 1))) {
+        result += String(delimiter);
+      } else if ((i === (this.length - 1)) && (this[i] !== undefined)) {
+        result += this[i];
+      } else if (i !== (this.length - 1)) {
+        result += this[i] + String(delimiter);
+      } else {
+        break;
```
polosanya: It looks too complicated. Consider optimizing your solution without 'else...if'
Add this[i] to the result if it's not null and it's not undefined.
And then add a separator if it's not the last element. That's all

polosanya (changes requested):
Nice! Let's try to improve your solution a bit)

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const delimiter = separator;
```
polosanya, line 9: This line is actually redundant, you can just use separator

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
polosanya, line 8: It is better to use more meaningful names, than just 'result'

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,22 @@
...
+    const delimiter = separator;
+
+    for (let i = 0; i < this.length; i++) {
+      if ((this[i] !== null) && (this[i] !== undefined)) {
+        result += this[i];
+      };
+
+      if (i !== (this.length - 1)) {
+        result += delimiter;
+      };
```
polosanya, line 18: 🔥

polosanya (approved):
Well done! I'm approving it, but please pay attention to the comments)

PR #2431, opened 2023-05-23
add task solution

w315:


src/arrayMethodJoin.js
```diff
@@ -3,9 +3,28 @@
...
+      } else if (this[i] === null || this[i] === undefined) {
+        str += separator;
+      } else if (separator === null) {
+        str += this[i] + 'null';
+      } else if (separator === undefined) {
+        str += this[i] + ',';
+      } else {
+        str += this[i] + separator;
+      }
+    }
```
vovakpro13: try to prettify it, get rid of 'if else' and move str += this[i] out of conditions as it's almost common command

vovakpro13 (changes requested):

Volodymyr-Melnyk-Fron10der (changes requested):
-

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,49 @@
...
+          continue;
+
+        case null:
+          str += this[i] + 'null';
+
+          continue;
+      }
+
+      str += this[i] + separator;
+    }
```
nazarmatsevych: Looks too complicated, actually we only need two checks to complete this task:

1. check if this[i] is not null and is not undefined then we can add this[i] to our str
2. check if current index is less than the last index so we can add separator to our str

And thats it

nazarmatsevych (changes requested):

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,26 @@
 /**
  * Implement method join
  */
+
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
nazarmatsevych, line 8:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

We can add a default value to separator

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,26 @@
...
+        str += this[i];
+      }
+
+      if (i < this.length - 1) {
+        if (separator === undefined) {
+          str += ',';
+        } else if (separator !== undefined) {
+          str += separator;
+        }
+      }
```
nazarmatsevych, line 22:
```suggestion
      if (i < this.length - 1) {
        str += separator;
      }
```

nazarmatsevych (approved):
Well done, 
left minor comment to improve your solution

PR #2433, opened 2023-05-25
Solution

Nikita211112:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let copy = '';
```
loralevitska, line 8: It is better to use a more descriptive name for the variable 'copy', like 'joinedString' for better code readability.

loralevitska (approved):
Good job!

PR #2437, opened 2023-05-26
Add join2

AmadeuszAndroid:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    if (separator === undefined) {
+      return this.join2(',');
+    }
```
DorotaLeniecDev: You can simplify using default parameter 
**function(separator = ',')**

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
...
+
+    let joinedString = '';
+    const arrayLength = this.length;
+
+    for (let e = 0; e < arrayLength; e++) {
```
DorotaLeniecDev: In most cases the convention is to name the iterator variable **i**

DorotaLeniecDev (changes requested):
Good job, two comments for improvements

src/arrayMethodJoin.js
AmadeuszAndroid: I moved it to default parameter.

src/arrayMethodJoin.js
AmadeuszAndroid: I renamed variable to _i_.

DorotaLeniecDev (approved):
Good job :)

PR #2438, opened 2023-05-26
Solution

kropiva:


src/arrayMethodJoin.js
```diff
@@ -1,11 +1,20 @@
...
- * Implement method join
- */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i !== this.length && i !== 0) {
```
TarasHoliuk: this `if` block should be after the next `if` block - it will be more natural (you are actually adding separator **after** each element) and will allow you to get rid of the `i !== 0` condition.

also, `i !== this.length` will never happen anyway, since your loop works while `i < this.length`, you shouldn't add separator after the last element (i.e. `this.length - 1`)

TarasHoliuk (changes requested):

TarasHoliuk (approved):

PR #2444, opened 2023-05-31
Solution

hulchak:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,28 @@
...
+          str += separator;
+        }
+      }
+    }
+
+    return str;
   };
 }
 
+// const source = [0, 1, 2, 3];
```
OleziO, line 28: Remove all comments from the code

OleziO (approved):

PR #2447, opened 2023-06-04
add solution

prostonadobobb:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        if (separator !== undefined) {
```
vlad-demchuk: you can set a default value for the argument to avoid this check

vlad-demchuk (changes requested):

OleziO (approved):

PR #2454, opened 2023-06-09
task solution

Bonum-taurus:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,40 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

We can set default value for separator

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,40 @@
...
+        return string;
+      }
+
+      const elem = this[i] + delimiter;
+
+      string += elem;
+    }
+
+    return string;
   };
```
nazarmatsevych: Solution looks too complicated, actually we need only two checks inside your loop:
1. Check if this[i] is not null and is not undefined  then we can add this[i] to our result string
2. Check if current index is not last index and then we can add delimiter to our result string and thats it

nazarmatsevych (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      const elem = this[i];
+
+      if (!exceptions.includes(this[i])) {
+        string += elem;
+      }
+
+      if (this[i] !== this[this.length - 1]) {
```
volodymyr-soltys97, line 23: Change the condition to `i !== this.length - 1` to avoid unnecessary array access and improve performance.

volodymyr-soltys97 (approved):
Nice work 👍

PR #2455, opened 2023-06-09
add task solution

MaxSchmide:


ollavka (approved):
Good job)

ollavka (approved):

YegorKorenkov (approved):
Looks good!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = this[0] === null ? '' : `${this[0]}`;
+
+    const separate = separator === null ? 'null' : separator;
```
tetiana-ivanchenko: redundant

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+      return string;
+    }
+
+    for (let i = 1; i < this.length; i++) {
+      let element = this[i];
+
+      if (this[i] === null || this[i] === undefined) {
+        element = '';
+      }
+      string += separate + element;
```
tetiana-ivanchenko: no need to add empty string, check if element is not null and undefined, and add it only in this case

tetiana-ivanchenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
+
+    const separate = `${separator}`;
```
anstsot: no need to add additional value, just use `separator`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      const element = this[i] === null || this[i] === undefined
+        ? ''
+        : this[i];
+
+      i === this.length - 1
+        ? resultString += element
+        : resultString += element + separate;
```
anstsot: try to rewrite by using `if ... else` construction or use the ternary operator after setting the value to `resultString`: `resultString += i === this.length - 1 ? ... : ....;`

anstsot (changes requested):
Almost done :+1:
Let’s add a few fixes to improve your code.

anstsot (approved):

PR #2456, opened 2023-06-09
Solution

rvrazovskyi:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,27 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const sep = typeof separator === 'undefined' ? ',' : separator;
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const element = this[i];
+
+      if (element === null || typeof element === 'undefined') {
```
VeronikaSynieokaia: Combine checks for null and undefined using 'element == null'

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,27 @@
...
+        if (i === this.length - 1) {
+          break;
```
VeronikaSynieokaia: Remove the 'break' statement, as it is not needed.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,27 @@
...
+        }
+        result += sep;
+      } else {
+        result += element;
```
VeronikaSynieokaia, line 22: Move appending 'sep' outside of the else block to reduce code duplication.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,27 @@
...
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const sep = typeof separator === 'undefined' ? ',' : separator;
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const element = this[i];
+
+      if (element === null || typeof element === 'undefined') {
+        if (i === this.length - 1) {
```
VeronikaSynieokaia: Remove unnecessary check for the last element, as it is handled by the loop condition.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,27 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const sep = typeof separator === 'undefined' ? ',' : separator;
```
VeronikaSynieokaia: Use '===' when checking for undefined: 'separator === undefined'

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,27 @@
...
+
+      if (element === null || typeof element === 'undefined') {
+        if (i === this.length - 1) {
+          break;
+        }
+        result += sep;
+      } else {
+        result += element;
+
+        if (i < this.length - 1) {
```
VeronikaSynieokaia: Move this check and appending 'sep' to the end of the loop, outside of the else block.

VeronikaSynieokaia (changes requested):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const sep = separator === undefined ? ',' : separator;
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const element = this[i];
+
+      if (element === null || element === undefined) {
```
PolinaRomanenko, line 14: The condition 'element === null || element === undefined' can be simplified to 'element == null'.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,30 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const sep = separator === undefined ? ',' : separator;
```
PolinaRomanenko, line 8: You should also handle the case when separator is null.

PolinaRomanenko (approved):
pay attention to the comments for the future)

PR #2458, opened 2023-06-09
Solution MR

RuslanMelnykFE:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+      if (arr[index] === undefined) {
+        str += `${separator}`;
+      } else if (arr[index] === 0 || isNaN(arr[index])) {
+        str += `${separator}${arr[index]}`;
+      } else {
+        str += `${separator}${arr[index] || ''}`;
+      }
+    }
+
+    return str;
```
nazarmatsevych: Good work but looks too complicated.
Actually we need only two checks inside the loop.

1. Check if this[i] is not null and is not undefined then you can add this[i] to result string.
2. Check if current index is not last index then you can separator to result string and thats it.

nazarmatsevych (changes requested):

VitaliyBondarenko1982 (approved):
Well done

PR #2462, opened 2023-06-11
solution

maxyma-sanx:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let resultingString = '';
+    const splitter = separator === undefined ? ',' : separator;
```
ilona-borkovska: you can set default value for parameter 'separator' like that: function(separator = ',') {} if function accepts nothing

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,21 @@
...
+
+    for (let i = 0; i < this.length; i += 1) {
```
ilona-borkovska, line 10:
```suggestion
    for (let i = 0; i < this.length; i++) {
```
use increment in this case

ilona-borkovska (approved):
Approve ✨

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
tetiana-ivanchenko: agree
https://github.com/mate-academy/js_array-method-join/pull/2462/files#r1225858779

src/arrayMethodJoin.js
tetiana-ivanchenko, line 10: https://github.com/mate-academy/js_array-method-join/pull/2462/files#r1225859032

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,21 @@
...
-    // write code here
+    let resultingString = '';
+    const splitter = separator === undefined ? ',' : separator;
+
+    for (let i = 0; i < this.length; i += 1) {
+      const item = this[i] === null || this[i] === undefined ? '' : this[i];
```
tetiana-ivanchenko: no need to add empty string, check if your item is not null and undefined, and add it only in this case

tetiana-ivanchenko (changes requested):

src/arrayMethodJoin.js
anstsot: as far as I understand you have eslint rule and can't use the increment, right?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let resultingString = '';
+
+    for (let i = 0; i < this.length; i += 1) {
+      let item = '';
+
+      if (this[i] !== null && this[i] !== undefined) {
+        item = this[i];
+      }
+      // const item = this[i] === null || this[i] === undefined ? "" : this[i];
```
anstsot: do not push the comments and try to keep the code clean

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+
+      if (i === this.length - 1) {
+        resultingString += `${item}`;
+        continue;
+      }
```
anstsot: can you merge these 2 conditions without creating additional `item` variable?
you can check the value is not `null` or `undefined` **and** not the last in array, then add `item` and `separator`

anstsot (changes requested):
Let’s add a few fixes to improve your code.

anstsot (approved):

PR #2466, opened 2023-06-11
add task solution

KandaskalovOleg:


irynabolbat (approved):
good job 👍

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
...
+    if (separator === undefined) {
+      newSeparator = ',';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if ((this[i] === undefined) || (this[i] === null)) {
+        arrToString += '';
+      } else {
+        arrToString += this[i];
+      }
```
```suggestion
      if (this[i] !== undefined && this[i] !== null) {
        arrToString += this[i];
      }
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let arrToString = '';
+    let newSeparator = separator;
+
+    if (separator === undefined) {
+      newSeparator = ',';
+    }
```
tetiana-ivanchenko: use default parametr instead

tetiana-ivanchenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let arrToString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        arrToString += '';
```
YegorKorenkov: Not fixed(previous review comment). Why do you need to do an extra check to add an empty string? This is not necessary, you already just add a separator on line 20, this is enough.

YegorKorenkov (changes requested):

YegorKorenkov (approved):
GJ!

PR #2469, opened 2023-06-12
Solution

Platon-jun:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,55 @@
...
+
+    if (this.length === 1) {
+      result += this[0];
+
+      return result;
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (i !== this.length - 1) {
+        switch (this[i]) {
```
volodymyr-soltys97: There is no need to check for undefined and null separately, as both cases have the same outcome. Use a simpler conditional statement.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,55 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    let modSeparator;
+
+    switch (separator) {
```
volodymyr-soltys97: Instead of using a switch statement for handling separator cases, consider using simpler conditional statements for better readability.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,55 @@
...
+            break;
+          case null:
+            result += modSeparator;
+            break;
+          default:
+            result += this[i] + modSeparator;
+            break;
+        }
+      } else {
+        switch (this[i]) {
```
volodymyr-soltys97: Remove the switch statement as it is not necessary here. Directly add the value of this[i] to the result if it is not undefined.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,42 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
Viktor-Kost: you can add default value to separator
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,42 @@
...
+
+    if (this.length === 0) {
+      return result;
+    }
+
+    if (this.length === 1) {
+      result += this[0];
+
+      return result;
+    }
```
Viktor-Kost: this part of code is redundant
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,42 @@
...
+        if (this[i] === undefined || this[i] === null) {
+          result += modSeparator;
+        } else {
+          result += this[i] + modSeparator;
+        }
+      } else {
+        if (this[i] !== undefined) {
+          result += this[i];
+        }
+      }
```
Viktor-Kost: we can simplify these conditions
```suggestion
        if (this[i] !== undefined && this[i] !== null) {
          result += this[i];
        } 
        
        if (i !== this.length - 1) {
          result += separator;
        }
```

Viktor-Kost (changes requested):

volodymyr-soltys97 (changes requested):
Not all fixed from previous review

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    let modSeparator = separator;
+
+    if (separator === null) {
+      modSeparator = 'null';
+    }
```
vlad-demchuk: you don't need this check

vlad-demchuk (changes requested):

src/arrayMethodJoin.js
Platon-jun: in the beginning my code was three times bigger. how to learn to make it like that right away

vlad-demchuk: it's okay, you will achieve it with practice
just keep doing

vlad-demchuk (approved):

PR #2473, opened 2023-06-12
Solution

dianadiv:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,31 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let dublicateSeparator = separator;
```
tetianakostiuk1: du<ins>**p**</ins>licate😊

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,31 @@
...
+    let joinedArray = '';
+
+    switch (separator) {
+      case undefined :
+        dublicateSeparator = ',';
+        break;
+
+      case null :
```
tetianakostiuk1: if I'm not mistaken the requirement is to use a default separator only when the parameter is undefined, and not when it's null, so there is no need to check for null 😊

tetianakostiuk1 (approved):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,31 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let dublicateSeparator = separator;
+    let joinedArray = '';
+
+    switch (separator) {
+      case undefined :
+        dublicateSeparator = ',';
```
tetiana-ivanchenko: use default parametr instead

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,31 @@
...
+        break;
+
+      case null :
+        dublicateSeparator = 'null';
+        break;
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null
+        && typeof this[i] !== 'undefined') {
```
tetiana-ivanchenko: why typeof? js hasn't "undefined" type

tetiana-ivanchenko (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedArray = '';
```
anstsot, line 8: better to name as `joinedString`

anstsot (approved):

PR #2474, opened 2023-06-12
Solution

yehor-kor:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
polosanya, line 8: It is better to avoid naming variables 'result'. Try to specify what exactly the result is (e.g. joinedString)

polosanya (approved):
Nice!

PR #2478, opened 2023-06-13
Solution

MariiaMonakhova:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
melnyksergey7: it will be better naming variable more correctly to understand, which action you need to do

f.e. let resJoined

melnyksergey7 (approved):
great work, but pay attention to the comment

tetiana-ivanchenko (approved):

PR #2483, opened 2023-06-13
code

DanielLinker17:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
SanyaBratashchuk:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        result += this[i];
      } 
```

SanyaBratashchuk (changes requested):

src/arrayMethodJoin.js
DanielLinker17: dosen't pass tests after this change

DarkMistyRoom (approved):
Good job!

PR #2484, opened 2023-06-13
Solution

olena-ole:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+    const glue = separator === null ? 'null' : separator;
```
MarianVolytskyi: if i understand, we never have "null", because we have  separator = ','

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+      const currentString = this[i] === null || this[i] === undefined
+        ? ''
+        : this[i];
+
+      if (i === this.length - 1) {
+        return joinedString + currentString;
+      }
+
+      joinedString += currentString + glue;
+    }
```
MarianVolytskyi: dont need to crate new variable(currentString), you can add current element to already created variable "joinedString", then check if it not last element and add separator.

MarianVolytskyi (changes requested):
Please read comments and optimize code.

src/arrayMethodJoin.js
NataliGru: I think this line `const glue = separator === null ? 'null' : separator;`
is unnecessary because you have already set a default value for the separator above)

NataliGru (approved):
Good job✨
Please remove the line indicated in the comments)

MarianVolytskyi (approved):

ManoylovAC (approved):
One of the possible ways - good 👍

PR #2485, opened 2023-06-13
Solution

IhorVerba:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
```

OleziO (changes requested):

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,22 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        resultString += separator;
+      }
+
+      if (this[i] !== null && typeof this[i] !== 'undefined') {
+        resultString += this[i];
+      } else {
+        resultString += '';
```
volodymyr-soltys97, line 15: This line is redundant because an empty string is already implicitly added when concatenating a value to a string. You can remove this line.

volodymyr-soltys97 (approved):
Nice work 🚀

PR #2488, opened 2023-06-14
arrayMethodJoin solution

ivanbilotserkivskiy:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringFromArray = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i === this.length - 1 && typeof this[i] !== 'undefined') {
+        stringFromArray += `${this[i]}`;
```
outofs: Well done, but no need to wrap element in backticks If you adding it to string. You can just use  `stringFromArray += this[i]`.

outofs (approved):
GW! Aproove!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+
+        return stringFromArray;
+      } else if (i === this.length - 1) {
+        return stringFromArray;
+      }
```
And678: No need to make checks for last element and return. Your `for` loop will end after last element. You have the return in the bottom of your function. Also, you can handle your comma logic once below, no need for special cases

```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+
+      if (typeof this[i] !== 'undefined' && typeof this[i] !== 'object') {
```
And678: We need to check for nulls here, not objects. Your join works incorrectly because of this - `[{}, {}, 1].join2(',')` returns different result compared to real `.join`.
```suggestion
      if (typeof this[i] !== 'undefined' && this[i] !== null) {
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+        stringFromArray += `${this[i]}${separator}`;
+        continue;
+      }
+      stringFromArray += `${separator}`;
+    }
```
And678: Avoid using `break`/`continue` unless not avoidable, it make code really hard to read. Again, we can handle separators once, no need to add it to every case

```suggestion
      if (typeof this[i] !== 'undefined' && typeof this[i] !== 'object') {
        stringFromArray += `${this[i]}`;
      }
      if (i !== this.length - 1) {
        stringFromArray += `${separator}`;
      }
    }
```

And678 (changes requested):
Some improvements needed, overall works ok, except of objects.

ManoylovAC (approved):
Good work 👍

PR #2491, opened 2023-06-14
Solution

outofs:


vladmakohon (approved):
Good solution!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let correctSeparator = separator;
+
+    if (separator === null) {
+      correctSeparator = 'null';
+    }
```
ManoylovAC: when you concatenate `string` with `null` or `undefined` you will get `string` so no need to do this additional manipulations
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let correctSeparator = separator;
```
ManoylovAC: you can simplify code and use only `separator` without `correctSeparator` - see other comments with explanations

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+
+      if (element === undefined || element === null) {
+        element = '';
+      }
+
+      if (i === this.length - 1) {
+        newString += element;
+      } else {
+        newString += element + correctSeparator;
+      }
```
ManoylovAC: not critical, but this part can be simplified
hint: you can always add elements if they exist, and only add a delimiter if it's not the last element

ManoylovAC (changes requested):
Good job
I left some comments on how you can quickly simplify your solution and not use variables that you really don't need

juljaVysotska (approved):
Good job!

PR #2492, opened 2023-06-14
add task solution

ArtemAndrushchenko:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let finalSeparator = separator;
+
+    if (finalSeparator === undefined) {
+      finalSeparator = ',';
+    }
```
cebanoleksandr: It would be better to set a default value, instead of this check. For example function(separator = ', ')

cebanoleksandr (changes requested):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,20 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const finalSeparator = separator === undefined ? ',' : separator;
```
anstsot: please add default value to `separator` param

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,20 @@
...
+    let result = '';
```
anstsot: better to rename to
```suggestion
    let joinedString = '';
```

anstsot (changes requested):
Almost done :+1:
Let's improve a little

tetiana-ivanchenko (approved):

PR #2495, opened 2023-06-14
Add task solution

vladmakohon:


ovynarchuk (approved):
good job! approved

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let resultStr = '';
+    let normalizedSeparator;
+
+    if (separator === undefined) {
+      normalizedSeparator = ',';
```
ManoylovAC: you can use it as default value for `separator` parameter

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
...
+    } else {
+      normalizedSeparator = String(separator);
```
ManoylovAC: all concatenations with string will give you `string' result so you can avoid this part as well

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
...
+      }
+
+      resultStr += String(value) + normalizedSeparator;
+    }
+
+    if (normalizedSeparator === '') {
+      return resultStr;
+    }
+
+    return resultStr.slice(0, -normalizedSeparator.length);
```
ManoylovAC: this is an unnecessary complication - please rethink your solution without the `slice` method

you can create your `resultStr` without this code-line by adding some extra condition to your previous code and return just `resultStr`

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,28 @@
...
+      if (value === null || value === undefined) {
+        value = '';
+      }
+
+      resultStr += String(value) + normalizedSeparator;
+    }
+
+    if (normalizedSeparator === '') {
+      return resultStr;
+    }
```
ManoylovAC: you can avoid this

ManoylovAC (changes requested):
Good start
And good time to refactoring)
Please look at my comments and rework some of your unnecessary complications.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+    let resultStr = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let value = this[i];
+
+      if (value === null || value === undefined) {
+        value = '';
+      }
+
+      resultStr += value;
```
ManoylovAC, line 17: you can additionally simplify this part and concatenate string only  if it not `null` and not `undefined`

ManoylovAC (approved):
Much better 👍 
Also left a comment for possible improvement

PR #2496, opened 2023-06-14
Solution

kylyvnyk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedStr = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        joinedStr += '';
+      } else {
+        joinedStr += this[i];
```
FlameInBrains: Great start. Only one little change is necessary
Here you can use one "if" operator in which you can write joinedStr += this[i];
It is not necessary to create an "if else" structure since the program will ignore 'null' and 'undefined' if the condition is written correctly;

FlameInBrains (changes requested):

ManoylovAC (approved):
Good work 👍

PR #2498, opened 2023-06-14
solution

AnatoliiLykhohodenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringifiedArr = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (typeof this[i] !== 'undefined' && this[i] !== null) {
```
vlopatko, line 11: But why you use "typeof" for undefined but don't use for null?

vlopatko (approved):
Good job, this code works. 👍

src/arrayMethodJoin.js
ManoylovAC: you don't need the `typeof` operator if you're just comparing to `undefined`
```suggestion
      if (this[i] !== undefined && this[i] !== null) {
```

ManoylovAC (approved):
Good work!
Тext time, please be careful about unnecessary operators

PR #2500, opened 2023-06-14
Finished task

IgorOksentyuk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const newArray = [];
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        newArray[i] = '';
+      } else {
+        newArray[i] = this[i];
```
NazarBaraban: [CODE STYLE] - avoid using nested if statements, we are sure that you can do it without them 
Use for example ternary operator like 2 strings lower or try to find another condition for `If`

NazarBaraban (changes requested):
Good job, but try to fix one part

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const newArray = [];
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      this[i] === null || this[i] === undefined
```
YegorKorenkov: Please try to avoid ternary operator, please try to use **if**. Also it's redundant to make newArray[i] = '' - empty string. If the condition is met, you can push this[i], if i !== this.length - 1   -  push separator.
1. if (this[i] !== null || this[i] !== undefined) - here it is enough to simply add an element to the term if it is not equal to null or undefined, else is not needed because why would we add an empty string when we can just add a separator

2. same with the second separator check, it is enough to use only if without else

3. in principle, we do not need an additional array, because you will not mutate the input, but only take each value, perform a check, and if it passed, then add it to the string str

Example for first condition:
```suggestion
      if (this[i] !== null || this[i] !== undefined) {
        str += this[i]
      }
```

YegorKorenkov (changes requested):
Almost done! Please check comments

juljaVysotska (approved):
Good job!

PR #2509, opened 2023-06-14
task solution

AlyonaSarapina:


kochetov-yevhen-may29 (approved):
Good solution!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
```
anstsot, line 8: better to rename to smth more meaningfull 
```suggestion
    let joinedString = '';
```

anstsot (approved):

PR #2514, opened 2023-06-14
implement join metod

kochetov-yevhen-may29:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let strFromArr = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const lastCheckSeparator = this[this.length - 1] === undefined
+      || this[this.length - 1] === null;
```
anstsot: why do you check pre last item? you already have the same check on line 14

anstsot (changes requested):
the code is working but looks overcomplicated. try to reduce it

src/arrayMethodJoin.js
kochetov-yevhen-may29: one of the tests crashed, when array has undifined at last element

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        strFromArr += this[i];
+      }
+
+      if (i !== this.length - 1) {
+        strFromArr += `${separator}`;
+        continue;
+      }
```
tetiana-ivanchenko, line 18:
```suggestion
      if (i !== this.length - 1) {
        strFromArr += `${separator}`;
      }
```

tetiana-ivanchenko (approved):

PR #2516, opened 2023-06-15
Implement applyCustomJoin function

taras-bilas:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+
+    for (let element = 0; element < this.length; element++) {
+      if (typeof this[element] !== 'undefined' && this[element] !== null) {
```
anstsot, line 11: can be simplified
```suggestion
      if (this[element] !== undefined && this[element] !== null) {
```

anstsot (approved):

PR #2518, opened 2023-06-15
Solution

NadiaKovalchuk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let separatedArr = '';
```
NataliGru: The`join()` method creates and returns a new string by concatenating all of the elements in an array, separated by commas or a specified separator string.
 - You are creating a string, not an array.
 - You are concatenating array elements, not separating them. So, please change the name)

NataliGru (changes requested):
Looks good! ✨✨✨
Please pay attention to the naming✨

tetiana-ivanchenko (approved):

PR #2520, opened 2023-06-15
add solution

akozlovska:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+      if ((this[i] === null || this[i] === undefined)
+        && i === this.length - 1) {
+        continue;
+      } else if (this[i] === null || this[i] === undefined) {
+        joined += separator;
+      } else if (i === this.length - 1) {
+        joined += this[i];
+      } else {
+        joined = joined + this[i] + separator;
+      }
```
sTorba24: can be simplified to these 2 actions
first - check if this[i] has value and add it
second - add separator excluding last index

sTorba24 (changes requested):
Almost good! Lets improve :)

andriy-shymkiv (approved):
verry good job!

PR #2525, opened 2023-06-15
Solution

s3rg4sh:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (this.length < 1) {
+      return '';
+    }
```
```suggestion
```
redundant

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        i === 0 ? result += '' : result += `${separator}`;
+      } else {
+        i === 0 ? result += `${this[i]}` : result += `${separator}${this[i]}`;
+      }
```
VitaliyBondarenko1982: let's improve readibility and amount of code.
avoid if else and ternary
it can be just two if
for make it try to use such checks for example:
`this[i] !== undefined && this[i] !== null`
and `i !== this.length - 1`

VitaliyBondarenko1982 (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        result += `${this[i]}`;
+      }
+
+      if (i !== this.length - 1) {
+        result += `${separator}`;
```
etojeDenys, line 16: The `continue` statement is unnecessary, as it will automatically move to the next iteration without it.

etojeDenys (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
```
OleziO, line 5: Remove all comments from the code.

OleziO (approved):

PR #2531, opened 2023-06-18
Solution

dipinchuk:


src/arrayMethodJoin.js
```diff
@@ -1,11 +1,28 @@
 'use strict';
 
-/**
- * Implement method join
- */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
tetiana-ivanchenko: use default parametr

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,28 @@
...
+
+    if (separator === null) {
+      addition = 'null';
+    }
+
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        str += i + 1 !== this.length ? this[i] + addition : this[i];
```
```suggestion
        str += this[i];
```

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,28 @@
...
+      } else {
+        str += i + 1 !== this.length ? addition : '';
+      }
```
```suggestion
      } 
      
      if (i !== this.length - 1) {
        str += separator;
      }
```

tetiana-ivanchenko (changes requested):

src/arrayMethodJoin.js
anstsot: the default param is still not fixed. please set the default value for `separator`

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,32 @@
...
-    // write code here
+    let addition = separator;
+
+    if (arguments.length === 0 || separator === undefined) {
+      addition = ',';
+    }
+
+    if (separator === null) {
+      addition = 'null';
+    }
```
anstsot: you don't have to introduce new `addition` variable. just use `separator`

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,32 @@
...
+
+    let str = '';
```
anstsot: try to use better namings as `joinedString`

anstsot (changes requested):
Almost done :+1:
Let's add few more improvements

anstsot (approved):

PR #2533, opened 2023-06-18
Solution

IrynaMP:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (typeof this === 'string') {
+      this.split(' ');
```
mykhalenych: You should remove the split method call, as join method should be applied only to arrays.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (typeof this === 'string') {
```
mykhalenych: You should remove the check for the string type, as join method should be applied only to arrays.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+      this.split(' ');
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
```
mykhalenych, line 12: You can remove this line as it's unnecessary to append an empty string. The else block is sufficient to handle non-null and non-undefined values.

mykhalenych (changes requested):

src/arrayMethodJoin.js
anastasiiavorobiova: Consider fixing the previous comment.

anastasiiavorobiova (changes requested):
Great solution! Added a tiny comment to fix

PR #2539, opened 2023-06-22
Solution

Luk2asz:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
DorotaLeniecDev: Use default parameter for separator
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
-    // write code here
+    let result = String(this[0]);
```
DorotaLeniecDev: You can silmy write
```suggestion
    let result = '';
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
+    let separatorVariable = separator;
```
DorotaLeniecDev: You dont need to create additional variable for separator, sipmly use **separator**

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,33 @@
...
+    }
+
+    const thisCopy = [...this];
+
+    for (let i = 1; i < this.length; i++) {
+      thisCopy[this.indexOf(undefined, i)] = '';
+      thisCopy[this.indexOf(null, i)] = '';
+
+      result += String(separatorVariable) + (thisCopy[i]);
+    }
```
DorotaLeniecDev: You can simplyfy your code, in for loop you can check if **this[i]** !== undefined oraz null and only then assign to **result**  value of **this[i]**, next you can assign **separator**, but you need to ommit this step for the last element
```suggestion
    for (let i = 1; i < this.length; i++) {
      //if (this[i] not equal to undefined and null) {
      //   result += this[i];
      //  }
      // if(condition when to add separator) {
      //  result += separator;
      // }
    }
```

DorotaLeniecDev (changes requested):
Code need some adjustments :)

darokrk (approved):
Looks nice 🥇  Good job 👍🏽

PR #2541, opened 2023-06-23
join

roman-tarasov90:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,18 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const result = [...this].toString();
+
+    switch (separator) {
+      case null:
+        return result.replaceAll(',', 'null');
```
VitaliyBondarenko1982: do not use other methods for implementing custom method.
just base js tools (`for loop`, `if`...)

VitaliyBondarenko1982 (changes requested):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,20 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
IvanFesenko: Separator could be any, not only comma, comma should be by dafult
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,20 @@
...
-    // write code here
+    const sourceStr = [...this].toString();
```
IvanFesenko: you should work with array items, not string
```suggestion

```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,20 @@
...
+
+    for (let i = 0; i < sourceStr.length; i++) {
+      if (sourceStr[i] === ',') {
+        result += separator;
+      } else if (separator === undefined) {
+        return sourceStr;
+      } else {
+        result += sourceStr[i];
+      }
+    }
```
IvanFesenko: Do not use else statement

```suggestion
    for (let i = 0; i < this.length; i++) {
    
      // Add each item to result,

       // if item is not last add separator
    }
```

IvanFesenko (changes requested):

mykhalenych (approved):

PR #2542, opened 2023-06-23
Solution AB

AytajBabayeva:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,21 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
     // write code here
+    const sep = separator !== undefined ? separator : ',';
```
witflash: You can use default assignment to function argument, I think.

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,21 @@
...
+    return this.reduce((acc, curr, index) => {
+      let currElement = curr;
+
+      if (currElement === null || currElement === undefined) {
+        currElement = '';
+      }
+
+      if (index === this.length - 1) {
+        return acc + String(currElement);
+      } else {
```
witflash: There is no need to use `else` here because you did return inside your `if` block. So you can just return anything directly after `if` block and this return will work only if your `if` condition did'n match.

witflash (changes requested):
Could your check my comments, please?

PR #2543, opened 2023-06-24
Solution

Patryk-Filipiak:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,16 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const sep = separator === undefined
+      ? ',' : separator === null ? 'null' : separator.toString();
```
JoyCoffeeAddict: Try to do a simple check, what happens, if you: 

```js
const exapmleArray = ['hello', 'world']
console.log(exampleArray.join(null))
console.log(exampleArray.join2(null))
```

You should not treat `null` differently than `undefined` here

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,16 @@
...
+    let str = '';
```
JoyCoffeeAddict: Choose more descriptive variable name.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,16 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      str = str + sep + (this[i] === null || this[i] === undefined
+        ? '' : this[i]);
```
JoyCoffeeAddict:
```suggestion
    if(this[i] != null ) { // non-strict equality for null and undefined
      str += sep + this[i]);
    }  
```

You can use ifs, but don't nest them unnecessarily
Also, checklist point 3

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,16 @@
...
+    }
+
+    return str.substring(sep.length, str.length);
```
JoyCoffeeAddict: I would create a variable then return it, it would be more readable.

JoyCoffeeAddict (changes requested):

src/arrayMethodJoin.js
JoyCoffeeAddict: Also, you don't have to specify  the second parameter of `.str.substring` (end index) since default means last index of a string already

src/arrayMethodJoin.js
Patryk-Filipiak: Hey, you are wrong in your suggested code it miss value, should put empty string... .
image

src/arrayMethodJoin.js
Patryk-Filipiak: done

src/arrayMethodJoin.js
Patryk-Filipiak: done

src/arrayMethodJoin.js
Patryk-Filipiak: i just changed to:
const separator = String(sep);

src/arrayMethodJoin.js
JoyCoffeeAddict: You can try this:

``` js 
	if (this[i] != null) {
	  joinString += separator + this[i];
	  continue;
	}
	joinString += separator + "";
```

Patryk-Filipiak: > 

works fine :) thanks and updated :)

JoyCoffeeAddict (approved):

PR #2544, opened 2023-06-24
Solution

Sirius9312:
The task 'Method join'

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -7,21 +7,12 @@ function applyCustomJoin() {
   [].__proto__.join2 = function(separator = ',') {
     let result = '';
 
-    if (separator === null) {
-      for (let i = 0; i < this.length; i++) {
-        result += (i !== this.length - 1)
-          ? this[i] + String(separator) : this[i];
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        result += this[i];
       }
-    } else {
-      for (let i = 0; i < this.length; i++) {
-        if (this[i] !== null && this[i] !== undefined) {
-          result += (i !== this.length - 1)
-            ? this[i] + separator : this[i];
-        } else {
-          result += (i !== this.length - 1)
-            ? '' + separator : '';
-        }
-      }
+
+      result += (i !== this.length - 1) ? separator : '';
     }
 
     return result;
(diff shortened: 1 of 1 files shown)


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,81 @@
  * Implement method join
  */
 function applyCustomJoin() {
+  let result = '';
+
   [].__proto__.join2 = function(separator) {
```
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```
just add default separator when it is not passed

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,81 @@
...
+            result += (i !== this.length - 1) ? '' + separator : '';
+          }
+        }
+        break;
+    }
+
+    return result;
   };
-}
+
```
VitaliyBondarenko1982: very long and complicated solution
it is possible to make much shorter
- make empty result string before for loop
- make for loop where in each iteration will be add item of this and separator (except last item) to result
- in loop can be just 2 if checks: 
   1. item of array not undefined and not null
   2. it is not last item

VitaliyBondarenko1982 (changes requested):
Try to make shorter solution

Sirius9312:
Fixed

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    if (separator === null) {
+      for (let i = 0; i < this.length; i++) {
```
volodymyr-soltys97: There is a redundant code duplication inside both loops, which can be avoided by moving the condition related to the last element after the loop.

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,28 @@
...
+        result += (i !== this.length - 1)
+          ? this[i] + String(separator) : this[i];
+      }
+    } else {
+      for (let i = 0; i < this.length; i++) {
```
volodymyr-soltys97: The checking of this[i] against null and undefined is not necessary as it's already done implicitly during string concatenation.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

Sirius9312:
Fixed-2

volodymyr-soltys97 (approved):
Awesome 🔥

PR #2547, opened 2023-06-27
add solution

tarovyk:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,27 @@
...
+    let result = '';
+    const length = this.length;
+    const lastElementIndex = length - 1;
+
+    for (let i = 0; i < length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
+      }
```
DarkMistyRoom:
```suggestion
      if (this[i] !== null && this[I] !== undefined) {
        result += this[i];
      }
```

DarkMistyRoom: `result += ''` is redundant

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,27 @@
...
+
+      if (i !== lastElementIndex) {
+        if (separator === undefined) {
+          result += ',';
```
DarkMistyRoom: use the default value for the separator instead of this condition

DarkMistyRoom (changes requested):
Good job! Just let's improve code a bit

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    const length = this.length;
+    const lastElementIndex = length - 1;
+    const defSeparator = ',';
```
SerhiiKirik: why do you need it?
You can just do like that
```js
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
...
+
+    for (let i = 0; i < length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        result += this[i];
+      }
+
+      if (i !== lastElementIndex) {
+        if (separator === undefined) {
+          result += defSeparator;
+        } else {
```
SerhiiKirik: try to avoid `if / else` construction

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
...
+    const lastElementIndex = length - 1;
+    const defSeparator = ',';
+
+    for (let i = 0; i < length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        result += this[i];
+      }
+
+      if (i !== lastElementIndex) {
+        if (separator === undefined) {
```
SerhiiKirik: remove it

SerhiiKirik (changes requested):
Great 🔥 
Let's make some improvements 🧑‍💻

Anastasiia-Svintsova (approved):

PR #2548, opened 2023-06-27
Solution array method join

Tvinki228:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
mykhalenych, line 7: Set a default value for 'separator' parameter to ','.

mykhalenych (approved):

PR #2551, opened 2023-06-28
Make tasks

F-u-n-H-o-u-s-e:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
```
OleziO: Remove all comments.

OleziO (changes requested):

demetrxx (approved):
Well done 👍 Go ahead!

PR #2552, opened 2023-06-28
solution

petylep:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let finalString = '';
+
+    if ((this[0] !== null)
+        && (this[0] !== undefined)) {
+      finalString += this[0];
+    }
```
```suggestion
```
looks like redundant if.
you can start from 0 in loop
and check i !== 0 for add separator

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+
+    for (let i = 1; i < this.length; i++) {
+      finalString += String(separator);
```
```suggestion
    for (let i = 0; i < this.length; i++) {
     if (i) {
        finalString += String(separator);
     }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+
+      if ((this[i] === null)
+          || (this[i] === undefined)) {
+        continue;
+      }
+
+      finalString += this[i];
```
```suggestion
      if (this[i] !== null  && this[i] !== undefined) {
       finalString += this[i];
      }
```
consider such refactor for increase readibility

VitaliyBondarenko1982 (changes requested):
Good job.
Consider a little refactoring

src/arrayMethodJoin.js
petylep: Ok, but I think this isn`t better solution: we will do extra ckeck avery loop.

VitaliyBondarenko1982 (approved):

PR #2553, opened 2023-06-28
Solution

AnnaViolin23:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i !== 0 && separator !== undefined) {
+        result += separator;
+      }
+
+      if (this[i] === null || this[i] === undefined) {
```
volodymyr-soltys97, line 15: You can simplify your code by using a ternary operator. Instead of using if/else, consider writing 'result += this[i] === null || this[i] === undefined ? '' : this[i]'.

volodymyr-soltys97 (approved):
Nice work 👍

PR #2556, opened 2023-06-29
add join method solution

vetal-hovenko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+      if (this[i] === null || this[i] === undefined) {
+        temp = '';
+      }
+
+      if (i !== lastIndex) {
+        separatedString += temp + String(separator);
+      } else {
+        separatedString += temp;
+      }
+    }
```
IvanFesenko: No need to reassign value, just add to result if it has value,
also add separator if i less then lastIndex
```suggestion
    for (let i = 0; i < this.length; i++) {
      const result = this[i];

      if (result !== null && result  !== undefined) {
        separatedString += result;
      }

      if (i < lastIndex) {
        separatedString += separator;
      }
    }
```

IvanFesenko (changes requested):
Good job. Let's improve your code a bit

IvanFesenko (approved):
great

PR #2557, opened 2023-07-01
solution

NataliMax:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,46 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
VitaliyBondarenko1982, line 7: set default value for separator as
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,46 @@
...
+
+        default:
+          if (this[i] === null || this[i] === undefined) {
+            result += separator;
+            break;
+          }
+          result += separator;
+          result += this[i];
+      }
+    }
```
VitaliyBondarenko1982, line 45: try to simplify your solution.
switch cases are redundant.
enough to make 2 checks inside loop
1. this[i] not equal null and this[i] not equal undefined
2. it is not last element.

VitaliyBondarenko1982 (changes requested):
Consider to make solution shorter.

VitaliyBondarenko1982 (approved):
Something went wrong with commit in PR, but see fixes and right solution in your account repo
Screenshot from 2023-07-02 14-44-03

well done

PR #2562, opened 2023-07-02
add solution

denys-kononenko:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      result += this[i] !== undefined && this[i] !== null
+        ? `${this[i]}`
+        : '';
```
```suggestion
      if (this[i] !== undefined && this[i] !== null) {
        result += this[i];
      }
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,21 @@
...
+
+      if (i !== this.length - 1) {
+        result += separator !== undefined
+          ? separator
+          : ',';
```
```suggestion
        result += separator;
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,21 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

We can add default value for separator so no need to check it for undefined

nazarmatsevych (changes requested):
Good work,
pls check comment to improve your solution

OleziO (approved):

PR #2563, opened 2023-07-02
solution

monikaprochniewicz:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,20 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
     // write code here
+    const sep = separator !== undefined ? String(separator) : ',';
```
DorotaLeniecDev: Instead of this use default value for parameter https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

DorotaLeniecDev (changes requested):
Use default parameter for the function

DorotaLeniecDev (approved):
Good job!

PR #2564, opened 2023-07-03
Solution

misha-shevchuk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      let item = this[i];
+
+      if (item === null || item === undefined) {
+        item = '';
+      }
+
+      if (i < this.length - 1) {
+        result += `${item}${separator}`;
```
volodymyr-soltys97: You are appending the separator even when the next item may be undefined or null, which will result in trailing separators in the output string. You should ensure that the next item is not null or undefined before appending the separator.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      let item = this[i];
```
volodymyr-soltys97: Instead of checking if the item is null or undefined, you could check if it is not undefined. This will handle all falsy values except for the number 0 and empty strings, which should be included in the joined string.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

VitaliyBondarenko1982 (approved):

PR #2565, opened 2023-07-04
method join v1

oltarasov:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,34 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let newstr = '';
+    let newseparator = separator;
+
+    for (let i = 0; i < this.length; i++) {
+      if (this.length === 0) {
```
mykhalenych: The check for empty array is unnecessary inside the loop. It should be placed before the loop starts.

mykhalenych (changes requested):

volodymyr-soltys97 (approved):
Nice work 🔥

PR #2567, opened 2023-07-04
Solution

EvgenBilenkiy:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
polosanya, line 8: It is better to avoid names such as `result`. Try to specify what exactly the result is

polosanya (approved):
Well done!

PR #2570, opened 2023-07-06
method join

Orchibald:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
IvanFesenko: just add default value - no need in additional variable
```suggestion
  [].__proto__.join2 = function(separator = ",") {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
...
-    // write code here
+    if (this.length === 0) {
+      return '';
+    }
+
+    const sep = separator === undefined ? ',' : String(separator);
+
+    let result = this[0] === null ? '' : String(this[0]);
```
IvanFesenko:
```suggestion
    let result =''
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
...
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] === undefined
+        || this[i] === null) {
+        result += sep;
+      } else {
+        result += sep + String(this[i]);
+      }
+    }
```
IvanFesenko: So steps are next:
check if `this[i]` not equal `undefined` or `null` - add to result
check if `i` is not last - add separator
```suggestion
    for (let i = 0; i < this.length; i++) {
      const value = this[i];
      
      if (value !== undefined   && value !== null) {
        result += value;
      } 
      
      if (i !== this.length - 1) {
        result  += separator;
      }
     
    }
```

IvanFesenko (changes requested):
Good job, but cade could be simplified

mykhalenych (approved):

PR #2575, opened 2023-07-10
Develop

Illia-L:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,26 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const length = this.length;
+    let workingSeparator = '' + separator;
```
anastasiiavorobiova, line 9: If you want to make a string conversion, convert it this way ```String(separator)```

anastasiiavorobiova (approved):
Great solution!

PR #2581, opened 2023-07-11
solution

olya-girnyak:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
```

OleziO (changes requested):

polosanya (approved):
Nice!

PR #2582, opened 2023-07-12
Add custom join method task solution

Serhii-Dr:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,35 @@
...
+
+    if (this.length === 0) {
+      return resultString;
+    }
+
+    if (this.length === 1) {
+      resultString += this[0];
+
+      return resultString;
+    }
```
Viktor-Kost: the check for array length being 0 or 1 is not necessary. The loop in the code will handle this case correctly.
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,35 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (separator === '') {
```
Viktor-Kost: Checking if separator is empty inside the loop is inefficient. This check should be done before the loop to avoid unnecessary comparisons.
So just modify the whole loop body to the code below

You need only to checks
```
        resultString += this[i];
      }

      if (i < this.length - 1) {
        resultString += separator;
      }
```

Viktor-Kost (changes requested):

Viktor-Kost (approved):
Cool!

PR #2589, opened 2023-07-17
Solution

Roman-Bilichenko:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let res = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
```
anastasiiavorobiova: The check for undefined and null is not sufficient. The original join function also excludes empty elements (undefined, null, and empty strings). You need to add a check for empty strings too.

anastasiiavorobiova (changes requested):
Great work! Added a tiny comment to fix

anastasiiavorobiova (approved):
Awesome!

PR #2590, opened 2023-07-17
Develop

andrew17752985:


src/arrayMethodJoin.js
```diff
@@ -1,11 +1,28 @@
...
-    // write code here
+    let result = '';
+
+    const length = this.length;
+
+    for (let i = 0; i < length; i++) {
+      const currentElement = this[i];
+
+      if (i !== 0) {
+        result += separator !== undefined ? separator : ',';
```
TarasHoliuk: let's just define the default separator in the function arguments

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,28 @@
...
 
 /**
  * Implement method join
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+
+    const length = this.length;
```
TarasHoliuk: you use it once, so it seems redundant

TarasHoliuk (changes requested):

qoa11a (approved):

nazarmatsevych (approved):

PR #2598, opened 2023-07-20
solution

oshapkun:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
...
-    // write code here
+    const actualSeparator = separator === undefined ? ',' : separator;
+
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const element = this[i];
+
+      if (i > 0) {
+        result += actualSeparator;
```
volodymyr-soltys97: The separator should not be appended if the current element is the last one in the array. You should check if i < this.length - 1 before appending the separator.

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
...
+      }
+
+      if (element === null || element === undefined) {
+        result += '';
```
volodymyr-soltys97: There is no need to append an empty string to the result. This operation could be removed for performance and readability.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code perfect

mykhalenych (approved):

PR #2600, opened 2023-07-20
solution

Sancz0pansa:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    let sep = separator;
```
JoyCoffeeAddict: I would not create a separate variable `sep` since you already handle `undefined`: `separator = ','` . Also, name `sep` is not descriptive

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,31 @@
...
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
+        if (i === this.length - 1) {
+          break;
+        }
+        result += sep;
+        continue;
+      }
```
JoyCoffeeAddict: Checkliost point 4 - nested ifs.

JoyCoffeeAddict (changes requested):

JoyCoffeeAddict (approved):

PR #2601, opened 2023-07-20
made method join

ivan-baranovskyi:


src/arrayMethodJoin.js
```diff
@@ -1,11 +1,30 @@
...
 'use strict';
 
 /**
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const result = this.slice();
```
Viktor-Kost: this variable is redundant
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,30 @@
...
+    let joinedString = '';
+
+    for (let i = 0; i < result.length; i++) {
+      const value = result[i];
+
+      if (value === null || value === undefined) {
+        joinedString += '';
+      } else {
+        joinedString += value;
+      }
```
Viktor-Kost: you can simplify
```suggestion
      if (value !== null || value !== undefined) {
        joinedString += value;
      } 
```

Viktor-Kost (changes requested):
Nice!
Some moments can be improved

Viktor-Kost (approved):

PR #2602, opened 2023-07-20
solution

BilokonYaroslav:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    const join2 = [...this];
```
polosanya: image

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+    let result = '';
+
+    for (let i = 0; i < join2.length; i++) {
+      if (join2[i] === undefined || join2[i] === null) {
+        join2[i] = '';
+      }
```
polosanya: What's the purpose of changing the array elements? Just add an element to the result if it's not undefined or null

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+
+      if (i === 0) {
+        result += join2[i];
+      }
+
+      if (i > 0) {
+        result = result + separator + join2[i];
```
polosanya: What about the last iteration? You don't need a separator there

polosanya (changes requested):
Good job! Let's improve your solution a bit)

src/arrayMethodJoin.js
BilokonYaroslav: I need separator at the last iteration, but make vithout it in first

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    // const join2 = [...this];
```
VitaliyBondarenko1982, line 9:
```suggestion
```
remove useless comments

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+        }
+      }
+
+      if (i > 0) {
+        if (this[i] === undefined || this[i] === null) {
+          result = result + separator + '';
+        } else {
+          result = result + separator + this[i];
+        }
+      }
```
VitaliyBondarenko1982, line 27: try to avoid nested if and if/else statements.
you need just two if

- current item not equal undefined and not equal null - in this case add current item to result
- current item not last item - in this case ass separator to result.

VitaliyBondarenko1982 (changes requested):

PR #2605, opened 2023-07-22
added my solution

FurArt:
Hi, and have a nice day

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,8 +4,30 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let join2String = '';
+
+    if (this.length === 1) {
+      return `${this[0]}`;
+    }
+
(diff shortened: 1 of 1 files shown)


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let join2String = '';
+
+    if (this.length === 1) {
+      return `${this[0]}`;
+    }
+
+    if (this.length > 1) {
```
TarasHoliuk: redundant check if you handle separator adding in a different way

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+      join2String = `${this[0]}`;
+
+      if (this[0] === null || this[0] === undefined) {
```
TarasHoliuk: just start iterating from `0` index

just add a separator after an element always except the last element

TarasHoliuk (changes requested):

volodymyr-soltys97 (approved):
Well done 🔥

PR #2606, opened 2023-07-23
Solution

oksom:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const arrayLength = this.length;
+    let result = '';
+
+    for (let i = 0; i < arrayLength; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += this[i];
```
maxim2310, line 15: Use explicit conversion to string here to ensure the correct type of the value - result += String(this[i]);

maxim2310 (approved):
Good Job:blush:!!!!

PR #2611, opened 2023-07-25
all tests pass

aniashev:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,30 @@
...
+      }
+    }
+
+    for (let i = 0; i < validElementsList.length; i++) {
+      result += validElementsList[i];
+
+      if (i < validElementsList.length - 1) {
+        result += separator;
+      }
+    }
```
IvanFesenko: validElementsList - is redudant.
No need in 2 loops
```suggestion

    for (let i = 0; i < this.length; i++) {
      const element = this[i];

      if (element !== null && element !== undefined) {
        result += element ;
      }
      
      if (i < this.length - 1) {
        result += separator;
      }
    }
```

IvanFesenko (changes requested):
Good job. Lets improve your code

VitaliyBondarenko1982 (approved):
GJ!

PR #2614, opened 2023-07-26
solution

helenmarth:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
VitaliyBondarenko1982: set default value of separator
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
...
+      if (this.length === 1) {
+        return '1';
+      }
+
+      if (this[i] !== undefined && this[i] !== null) {
+        result += this[i];
+      }
+
+      if (i !== this.length - 1) {
+        result += separator === undefined ? ',' : separator;
```
VitaliyBondarenko1982: after setting default value of separator this ternary will be redundant
```suggestion
        result += separator;
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,23 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this.length === 1) {
+        return '1';
+      }
```
```suggestion
```
looks like redundant. and why return '1'?

VitaliyBondarenko1982 (changes requested):

IvanFesenko (approved):
Well done

PR #2615, opened 2023-07-26
Solution

HaniaNassalska:


AndrzejPec (approved):
I guess logic could be simplified but it seems to be fine apart from that

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,22 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+    const resultSeparator = separator !== undefined ? `${separator}` : ',';
```
JoyCoffeeAddict: how about using `String` here? You can read about it [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String?retiredLocale=pl)

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,22 @@
...
+      if (this[i] === null || this[i] === undefined) {
+        (i === this.length - 1)
+          ? result += ''
+          : result += '' + resultSeparator;
+      } else {
+        (i === this.length - 1)
+          ? result += this[i]
+          : result += this[i] + resultSeparator;
+      }
+    };
```
JoyCoffeeAddict: I'm 100% sure you can think of an idea to simplify it! In case of any problems tag me here.

JoyCoffeeAddict (changes requested):

HaniaNassalska (commented):
I saw a similar solution while doing a review for a colleague and I think it's the best and simplest 😉

asakevych (approved):
Great job!! 🚀 🔥 👏 
but next time please use more descriptive names for variables -> insted of `result` it can be `joinedString` or something to make reader understand what value can have your variable 😉

PR #2617, opened 2023-07-26
task has been done

vladkorobka:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
+
+    if (!this.length) {
+      return resultString;
+    }
+
+    for (const element of this) {
+      if (element !== null && element !== undefined) {
```
anastasiiavorobiova: The original join function also excludes empty elements (undefined, null, and empty strings). You need to add a check for empty strings too.

anastasiiavorobiova (changes requested):
Awesome work! Almost done &#128170;

anastasiiavorobiova (approved):
Awesome!

PR #2618, opened 2023-07-26
Solution

myroslavaKukhta:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+    const elementToString = (value) =>
+      value !== null && value !== undefined
+        ? value : '';
+
+    for (let i = 0; i < this.length - 1; i++) {
+      resultString += elementToString(this[i]);
+      resultString += separator;
+    }
+
+    resultString += elementToString(this[this.length - 1]);
```
maxim2310, line 19: This line will fail if the array is empty (length 0), because this[-1] is undefined. It's better to start the loop from 0 to this.length and only append the separator if the current index isn't the last one.

maxim2310 (changes requested):
Good Job!!!
Please fix this small mistake

PR #2622, opened 2023-07-27
Solution

klaudia-klis:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const arrLength = this.length;
+    let result = '';
+    let newSeparator = separator;
+
+    if (separator === undefined) {
+      newSeparator = ',';
+    }
```
AndrzejPec: I see that you are assignin separator to a new value (line 10) because you want to reasign it (line 13), but you can consider asigning a default value for the separator, like so:

= function(separator = ',') [line 7]

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
+    for (let i = 0; i < arrLength - 1; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '' + `${newSeparator}`;
+      } else {
+        result += this[i] + `${newSeparator}`;
+      }
+    }
+
+    if (this[arrLength - 1] === undefined) {
+      result += '';
```
AndrzejPec: I'm not quite sure if this operation actually changes the 'result' string in any way, but if so, maybe consider flipping the logic and just not handling the other case - should be less complicated :)

AndrzejPec (changes requested):
Pretty close to perfection ;)

choeqq (approved):
Great job ✅

PR #2624, opened 2023-07-27
Develop

Matylda-Lewandowska:
Task solution of array-method-join;

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -8,14 +8,12 @@ function applyCustomJoin() {
     let joinedString = '';
 
     for (let i = 0; i < this.length; i++) {
-      if (i === 0 && (this[i] === null || this[i] === undefined)) {
-        continue;
-      } else if (this[i] === null || this[i] === undefined) {
+      if ((this[i] !== null && this[i] !== undefined)) {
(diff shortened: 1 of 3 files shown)


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (i === 0 && (this[i] === null || this[i] === undefined)) {
+        continue;
+      } else if (this[i] === null || this[i] === undefined) {
+        joinedString += `${separator}`;
+      } else if (i === 0) {
+        joinedString = `${this[0]}`;
+      } else {
+        joinedString += `${separator}` + `${this[i]}`;
+      }
```
choeqq: those condition seems to be little complex, please try something like this:
```suggestion
      if ((this[i] !== null && this[i] !== undefined)) {
          // add letter to joinedString
        }
        
      if (iterator is less than word length) // add separator to joinedString
```

choeqq (changes requested):
Almost there, you can make it little simpler 😄

Matylda-Lewandowska:
I managed to make these adjustments, thanks, the code looks a lot better now! :D

choeqq (approved):
Great job ✅

PR #2625, opened 2023-07-27
solution

ogospadzipalina:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,25 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+
+    let separatorNew = separator;
+
+    if (separator === undefined) {
+      separatorNew = ',';
+    }
```
AndrzejPec: this same result can be neatly achieved by declaring a default value for function parameter (line 7):

function(separator = ',')

AndrzejPec (changes requested):
almost flawless 👍

src/arrayMethodJoin.js
ogospadzipalina: @AndrzejPec thank you! all changed 👍

choeqq (approved):
Great job ✅

PR #2626, opened 2023-07-27
Solution

YuraMisura:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,18 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      result += this[i] !== undefined && this[i] !== null ? this[i] : '';
```
Viktor-Kost, line 11: this line of code converts these values to empty strings rather than omitting them completely. 
You'd better just do nothing, if it's not pass a condition
```suggestion
      if (this[i] !== undefined && this[i] !== null) {
        result += this[i];
      }
```

Viktor-Kost (approved):
Nice!
Just check a comment for yourself

PR #2628, opened 2023-07-27
solution

kuzgan:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+      if (this[i] === null || this[i] === undefined) {
+        newString += separator;
+        continue;
+      }
+
+      if (i === 0) {
+        newString += this[i];
+        continue;
+      }
+      newString += `${separator}${this[i]}`;
```
choeqq: instead of handling all other cases, we can try to look for valid ones and then perform our operation, that way we can get rid of multiple continue statements:
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
           // add letter to newString
       }
      
      if(iterator is less than word length) // add sepatator to newString
```

choeqq (changes requested):
Almost there, your solution can be improved ✅

choeqq (approved):
Well done ✅

PR #2630, opened 2023-07-27
solution no1

cardinal1991:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,25 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
choeqq: you can set default separator to `','`

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,25 @@
...
+    for (let i = 0; i < this.length; i++) {
+      // get rid off null and undefined
+      if ((this[i] !== undefined) && (this[i] !== null)) {
+        result += this[i];
+      }
+
+      // separator
+      if ((i < this.length - 1) && (separator === undefined)) {
+        result += ',';
+      }
```
choeqq: if you set default separator, you will be able to omit that condition 🚀

choeqq (changes requested):
Good job, but you can make it little simpler 🚀

choeqq (approved):
Good job ✅

PR #2631, opened 2023-07-27
Solution

Vira-V:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
...
+        if (i !== 0) {
+          newString += separator;
+        }
+        continue;
+      }
+
+      if (i !== 0) {
+        newString += separator;
+      }
+      newString += this[i];
```
choeqq: I think it would be much easier to check if this[i] is correct and then perform operation instead of handling all different cases:
```suggestion
      if (this[i] !== null || this[i] !== undefined) {
          // add letter to newString
      }
      
      if(iterator is smaller than word length) // add separator
```

choeqq (changes requested):
Almost there - you can make it little simpler 🚀

asakevych (approved):
Great job!!! 🚀 🔥 👏

PR #2634, opened 2023-07-28
solution

SviatoslavVasyletskyi:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
```
choeqq, line 8: you can do little better on that name 😄

choeqq (approved):
Good job 🚀 

Please take a look at my comment

PR #2639, opened 2023-07-28
add task solution

MateuszSlezak:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,35 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let newString = '';
+    let sepanullator = separator;
```
choeqq: what's sepanullator? 😄

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,35 @@
...
+
+    if (separator === null) {
+      sepanullator = 'null';
+    }
```
choeqq: separator won't ever be null - we have defaultet it to `','`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,35 @@
...
+        newString += sepanullator;
+        continue;
+      }
+
+      if (i === this.length - 1) {
+        newString += this[i];
+        break;
+      }
+
+      newString += this[i] + sepanullator;
```
choeqq: It is way easier to check for cases when we want to apply logic, instead of getting rid of everyone single other one 🔽 
```suggestion
      if (this[i] !== null && this[i] !== undefined) {
        newString += String(this[i])
      }

      if (i < this.length - 1) {
        newString += separator
      }
```

choeqq (changes requested):
Almost there, small changes needed 🚀

choeqq (changes requested):
Waiting for any updates @MateuszSlezak 👀

src/arrayMethodJoin.js
MateuszSlezak: Ah was sure i need this for this test: test
source = [0, 1, 2, 3];

(`Should work with null`, () => {
  expect(source.join2(null))
    .toBe('0null1null2null3');
});

choeqq (approved):
Looks good to me, nicely done ✅

PR #2640, opened 2023-07-29
Add task solution

spojrzenie:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const finalSeparator = separator === undefined ? ',' : separator;
```
Radoslaw-Czerniawski: This could be removed by declaring a default value of the parameter and then just using the separator
```js
```

Radoslaw-Czerniawski (approved):
Good job :), left one comment regarding the unnecessary declaration of variable

PR #2642, opened 2023-07-29
add task solution

arturwisnia:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i !== 0) {
+        result += separator === undefined ? ',' : separator;
```
choeqq, line 12: you should probably keep both check in if statement but that also works 😄 
```suggestion
      if (i !== 0 && separator !== undefined) {
        result += separator
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,19 @@
...
+      }
+
+      if (this[i] !== null && this[i] !== undefined) {
```
choeqq, line 15: that's little better in my opinion 🚀

choeqq (approved):
Good job 🚀 

Please take a look at my comments 😉

PR #2643, opened 2023-07-29
Solution

emil-owczarek:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,36 @@
...
+
+    if (this.length === 0) {
+      return output;
+    }
+
+    if (this[0] !== undefined && this[0] !== null) {
+      output += `${this[0]}`;
+    } else {
+      output += '';
+    }
```
choeqq: all of that checking can be done inside loop

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,36 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
choeqq: it would be smart to add default separator - logic with actualSeparator wouldn't be needed 😄 
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,36 @@
...
+      output += `${this[0]}`;
+    } else {
+      output += '';
+    }
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        output += `${actualSeparator}${this[i]}`;
+      } else {
+        output += actualSeparator;
```
choeqq: I think we want to add separator only when `i < this.length - 1` 🤔

choeqq (changes requested):
Almost there, you can improve your solution 👍

mvjl000 (approved):
Good job! 🚀

choeqq (approved):

PR #2647, opened 2023-07-30
solution

patryk177m:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinArray = '';
+    let secondSeparator = separator;
+
+    if (separator === null) {
+      secondSeparator = 'null';
+    }
```
choeqq: we don't really need to check if separator is null - you've defaulted it to `,`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
```
choeqq, line 12: this check is enough to assing `this[i]` to result string

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+        if (i === this.length - 1) {
+          joinArray += this[i];
+        } else {
+          joinArray += this[i] + secondSeparator;
+        }
+      } else {
```
choeqq: I think those can be ommited, aside from last, which is necessary

choeqq (changes requested):
Almost there, you can improve your solution 🚀

mvjl000 (approved):

choeqq (approved):

PR #2648, opened 2023-07-30
Add join solution

ghost:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
```
choeqq, line 8: I think you can do better on that name 😄

choeqq (approved):
Nicely done, please name your variables little bit more descriptively 🚀

PR #2651, opened 2023-07-31
Method join

Kacper-Leman:
Soultion to tests

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -5,7 +5,34 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let separatorValue = separator;
+
+    if (separatorValue === undefined) {
+      separatorValue = ',';
+    }
+
+    const arr = this;
+    let result = '';
+
+    if (arr.length === 0) {
+      return '';
+    }
+
+    for (const element of arr) {
+      const str = element === null
+      || element === undefined ? '' : element.toString();
+
+      result += str;
+
+      if (arr.indexOf(element) < arr.length - 1) {
+        if (separatorValue === null) {
+          separatorValue = 'null';
+        }
+        result += separatorValue.toString();
+      }
+    }
+
+    return result;
   };
 }
 
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -4,31 +4,15 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    let separatorValue = separator;
-
-    if (separatorValue === undefined) {
-      separatorValue = ',';
-    }
-
-    const arr = this;
+  [].__proto__.join2 = function(separator = ',') {
     let result = '';
 
-    if (arr.length === 0) {
-      return '';
-    }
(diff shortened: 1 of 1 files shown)


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,34 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let separatorValue = separator;
```
Radoslaw-Czerniawski: You may use the `separator` parameter and get rid of this declaration along with if, if you add a default value in the parameter declaration
```js
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,34 @@
...
+
+    if (separatorValue === undefined) {
+      separatorValue = ',';
+    }
+
+    const arr = this;
```
Radoslaw-Czerniawski: I'd say `arr` variable is unnecessary, from the context of this implementation it's implied this refers to an array

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,34 @@
...
+    let result = '';
+
+    if (arr.length === 0) {
+      return '';
+    }
+
+    for (const element of arr) {
```
Radoslaw-Czerniawski: If you need an index regular for loop with `let i = 0; i < this.length; i++` would be better

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,34 @@
...
+      const str = element === null
+      || element === undefined ? '' : element.toString();
```
Radoslaw-Czerniawski: `element.toString();` -> regarding this part, I personally am not sure whether all data types have `.toString` method, it would be safer to user something like `String(element)` or  interpolate `${element}` 
```js
result += this[i] === null || this[i] === undefined ? '' : String(this[i])
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,34 @@
...
+
+      result += str;
+
+      if (arr.indexOf(element) < arr.length - 1) {
```
Radoslaw-Czerniawski: If you go with the previously suggested changes this here could be shortened to:
```js
if (i < this.length - 1) {
  result += String(separator)
}
```

Radoslaw-Czerniawski (changes requested):
Needs some refactoring

Kacper-Leman:
I refactor this code like you said and yea is shorter and more readable :D

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      result += this[i] === null || this[i] === undefined
+        ? '' : String(this[i]);
```
DorotaLeniecDev: In my opinion this would be even more readable:

```js
        result += String(this[i]);
      }
```

DorotaLeniecDev (approved):
Good job one suggestion from my side, for more readability, you can apply it 
Approving 💪

PR #2652, opened 2023-07-31
add task solution

atrofymovych:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,26 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+
+    let sep = separator;
+
+    if (sep === undefined) {
```
anastasiiavorobiova: Instead of checking if 'sep' is undefined, consider checking if 'separator' is undefined before assigning it to ','. This would eliminate the need for the 'sep' variable.

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,26 @@
...
+      sep = ',';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
```
anastasiiavorobiova: The original join function also excludes empty elements (undefined, null, and empty strings). You need to add a check for empty strings too.

anastasiiavorobiova (changes requested):
Great work! Added tiny comments to fix

volodymyr-soltys97 (approved):
Well done 🔥

PR #2656, opened 2023-08-01
solution

mpytlyk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    // Helper function to handle empty elements
```
loralevitska: Remove comment. Try it write obvious code that explain all we need without comments)

loralevitska (approved):
Good job!

src/arrayMethodJoin.js
mpytlyk: Ok, thank you)

PR #2659, opened 2023-08-01
Solution

Reject-4444:


src/arrayMethodJoin.js
```diff
@@ -3,9 +3,45 @@
...
+
+  return element;
+}
+
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const separate = separatorHandler(separator);
```
Esceype, line 30: Try to use native js and here you can use and check if this will work to your solution
```suggestion
    const separatorStr = String(separator);
```

Esceype (approved):
GJ!!! Here is some code that you can test maybe that can improve your code solution

PR #2661, opened 2023-08-01
solved task

AndriiKuz:


davinnchii (approved):
LGTM

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringSeparator = separator;
+    let result = '';
+
+    if (stringSeparator === null || typeof stringSeparator === 'object') {
+      stringSeparator = String(stringSeparator);
+    }
```
Moroz-Dmytro: Do you really need this?
Result will be the same with and without this check
image

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringSeparator = separator;
```
Moroz-Dmytro: We don't need this variable

Moroz-Dmytro (changes requested):
GJ 👍 
But let's make your code cleaner

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        result += separator;
+      }
+
+      const element = this[i] === null || this[i] === undefined ? '' : this[i];
```
tiserett: let's create variable for condition and for this[i]

tiserett (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        result += separator;
+      }
+
+      const item = this[i];
+      const isValid = item === null || item === undefined;
+      const element = isValid ? '' : item;
```
tiserett, line 17: We set empty string if isValid, looks a bit strange

tiserett (approved):

PR #2662, opened 2023-08-01
Solutions

SergiyVS:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,35 @@
...
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let joinedStr = '';
+    let separatorInn = separator;
+    const arrInn = [...this];
+
+    if (separator === undefined) {
+      separatorInn = ',';
+    }
```
maxim2310: add default separator like this instead of if block
[].__proto__.join2 = function(separator = ',')

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,35 @@
...
+
+    for (let i = 0; i < arrInn.length; i++) {
+      if (arrInn[i] === null || arrInn[i] === undefined) {
+        arrInn[i] = '';
+      }
+    }
+
+    if (arrInn.length === 0) {
+      return '';
+    }
```
maxim2310: this case you can move to start before creating variable and use this.length

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,35 @@
...
+
+    if (separator === undefined) {
+      separatorInn = ',';
+    }
+
+    for (let i = 0; i < arrInn.length; i++) {
+      if (arrInn[i] === null || arrInn[i] === undefined) {
+        arrInn[i] = '';
+      }
+    }
```
maxim2310: this loop you can combine with next one

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,35 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let joinedStr = '';
+    let separatorInn = separator;
+    const arrInn = [...this];
```
maxim2310: make naming more meaningful

maxim2310 (changes requested):
Good job!!!

Anastasiia-Svintsova (approved):

PR #2663, opened 2023-08-01
add task solution

Koliras:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
```
voievodik: It's a good point to avoid undefined values, but in my opinion probably it's unnecessary in this case. Everything works without checking

voievodik (changes requested):
Good job!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const STRING_SEPARATOR = String(separator);
+    let finalString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === undefined || this[i] === null) {
```
Moroz-Dmytro: Create a variable for this[i]. You are using this value 3 times

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+        finalString += STRING_SEPARATOR;
+        continue;
+      }
+
+      finalString += this[i] + STRING_SEPARATOR;
+    }
+
+    return finalString.slice(0, finalString.length - STRING_SEPARATOR.length);
```
Moroz-Dmytro: Try to simplify this logic. Better will be to add 1 more `if` to your loop and to add `separator` only when this[i] is not the last element

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const STRING_SEPARATOR = String(separator);
```
Moroz-Dmytro: I guess you don't really need this. It will work fine without converting it to string

Moroz-Dmytro (changes requested):
Good work 👍 
But let's simplify your solution

tiserett (approved):

PR #2665, opened 2023-08-02
Add task solution

davinnchii:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,37 @@
...
+    } else if (separator === null) {
+      separatorCopy = 'null';
+    }
+
+    for (let item = 0; item < this.length - 1; item++) {
+      // Change all ignored characters to separator(
+      // f.e undefined or empty string, but exclude false, null, NaN)
+      if (!this[item] && typeof this[item] !== 'number'
+      && this[item] !== false) {
+        thisCopy[item] = separatorCopy;
```
Wesses: I think u can do it in one line;
thisCopy[item] = separatorCopy;
returnString += thisCopy[item];
-> returnString += separatorCopy;

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,37 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
     // write code here
+    let returnString = '';
+    let separatorCopy = separator;
+    const thisCopy = [...this];
```
Wesses: I am not sure we need to copy array, because we don't mutate it.

Wesses (changes requested):
Well done.

Wesses (approved):
Well done

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,34 @@
...
+    for (let item = 0; item < this.length - 1; item++) {
+      // Change all ignored characters to separator(
+      // f.e undefined or empty string, but exclude false, null, NaN)
+      if (!this[item] && typeof this[item] !== 'number'
+      && this[item] !== false) {
+        returnString += separatorCopy;
+      } else {
+        returnString += this[item].toString() + separatorCopy;
+      }
+    }
```
tiserett: let's use 'i' instead of item
then create variable item - this[i]

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,34 @@
...
+
+    // Check if last character is valid and then add it to our string
+    if (this[this.length - 1]) {
+      returnString += this[this.length - 1];
```
tiserett: create separate value for last character

tiserett (changes requested):

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,37 @@
...
+    let separatorCopy = separator;
+    const thisLastChar = this[this.length - 1];
+    // Check if our separator is undefined or null,
+    // so we change it to value we need
+
+    if (separator === undefined) {
+      separatorCopy = ',';
+    } else if (separator === null) {
+      separatorCopy = 'null';
+    }
```
BudnikOleksii, line 19: Avoid the "else if" statement

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,37 @@
...
+      const item = this[i];
+      // Change all ignored characters to separator(
+      // f.e undefined or empty string, but exclude false, null, NaN)
+
+      if (!item && typeof item !== 'number'
+      && item !== false) {
+        returnString += separatorCopy;
+      } else {
+        returnString += item.toString() + separatorCopy;
+      }
```
BudnikOleksii, line 31: The same with else, you can use the "continue" operator in the "if" statement, otherwise do the operation after the "if" block.

BudnikOleksii (approved):
Good job!
Left some comments that can improve your code, check them. Also, don't leave comments, your code should describe what it does.

Anna-Tkachenko (approved):

PR #2670, opened 2023-08-02
add task solution

Vannnkof:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
+
+    if (this[0] !== null && this[0] !== undefined) {
+      resultString = `${this[0]}`;
```
ivannaromanovych: You don't need to use back quotes here

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+    }
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        resultString += `${separator}` + `${this[i]}`;
+        continue;
+      }
+
+      resultString += `${separator}`;
```
ivannaromanovych: Same with quotes here

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
+
+    if (this[0] !== null && this[0] !== undefined) {
+      resultString = `${this[0]}`;
+    }
+
+    for (let i = 1; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
```
ivannaromanovych, line 11: From CHECKLIST:

> [CODE STYLE] - avoid using nested `if` statements, we are sure that you can do it without them :)

ivannaromanovych (approved):

src/arrayMethodJoin.js
Moroz-Dmytro: Good point

src/arrayMethodJoin.js
Moroz-Dmytro: Also, do you really need to do this here for the first element? You have the same condition(`this[i] !== null && this[i] !== undefined`) inside the loop. Try to combine them(but need to change some logic), your code will be more clear

src/arrayMethodJoin.js
Moroz-Dmytro: Here it's okay. It's not a nested `if statement`. Example of nested if:
`if (...) {
  if(...) {
    if(...) {
      ...
    }
  }
}`

Moroz-Dmytro (changes requested):
Good progress 👍 
But try to make your solution clearer

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        resultString += `${this[i]}`;
```
BudnikOleksii, line 12: You don't need interpolation here as the previous interviewer mentioned.

BudnikOleksii (approved):
Good job!
Pay attention to suggestions.

PR #2672, opened 2023-08-02
task was done

LinaDo1609:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const separatorString = `${separator}`;
```
etojeDenys: you can use the separator directly

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,19 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        result += this[i] + separatorString;
+      } else {
+        result += separatorString;
+      }
```
```suggestion
      if (this[i] !== undefined && this[i] !== null) {
        result += this[i] + separatorString;
      } 
      
      if (i < this.length - 1) {
        result += separatorString;
      }
```

etojeDenys (changes requested):

VitaliyBodnar (approved):
Good job!

PR #2673, opened 2023-08-02
Solution

deandre25:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
+
+    let res = '';
```
denys-danyliuk:
```suggestion
    let joinedString = '';
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+
+    if (this[0] === null) {
+      res += ``;
+    } else {
+      res += `${this[0]}`;
+    }
+
```
denys-danyliuk: you can do this inside of cycle

denys-danyliuk (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    if (this.length === 0) {
+      return '';
+    }
```
xiimiik, line 10: this condition already exists in the loop condition

xiimiik (approved):
Approve, but take care about comment :)

PR #2674, opened 2023-08-03
feat: add task solution

FrontKid:


Yevheniia-Sid (approved):
Great work🚀

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,34 @@
...
  * Implement method join
  */
+
+const isNotIgnoredEl = (el) => ![null, undefined].includes(el);
+
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const currentArray = this;
```
Moroz-Dmytro: I guess we don't really need to create this variable

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,34 @@
...
+
+    if (!(currentArray instanceof Array)) {
+      return '';
+    }
```
Moroz-Dmytro: On line 10 you create this method as a method of `Array`. So this method can be called only on array

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,34 @@
...
+
+    const INDEX_OF_LAST_ELEMENT = currentArray.length - 1;
```
Moroz-Dmytro: If you want to create variable for this, name it in camelCase

Moroz-Dmytro (changes requested):
GJ 👍 
Small comments to improve your code

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,28 @@
...
+
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const indexOfLastElement = this.length - 1;
+
+    let concatedString = '';
+
+    for (let curIndex = 0; curIndex < this.length; curIndex++) {
```
tiserett, line 15: I feel like curIndex is a redundant name, because everyone knows that 'i' is current index ;)

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,28 @@
 /**
  * Implement method join
  */
+
+const isNotIgnoredEl = (el) => ![null, undefined].includes(el);
```
tiserett, line 7: let's check if element isIgnored, and then set value to opposite, in place where we call it, I guess that's gonna be more correct

tiserett (approved):

PR #2675, opened 2023-08-03
JS array method join

vserhiichuk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+    let valueSeparatore = separator;
+
+    if (separator === null) {
+      valueSeparatore = 'null';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      const arrayValue = this[i] === undefined
```
davinnchii: Here i'd like to name the variable something like arrayItem or arrayElement, cuz value is more used in some loops or parameters for functions.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
```
davinnchii: I'd think would be better to name this variable like returningString or something like that, for it to be obvious that this is a string which we will return as a result.

davinnchii (changes requested):
Very good, but I suggested u renaming variables to increase readability
Also I find writing comments before loops to be a good practice, so it will be easier for new people to understand what exactly your loop or expression is doing.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
+    if (separator === null) {
+      valueSeparatore = 'null';
+    }
+
+    for (let i = 0; i < this.length; i++) {
+      const part = this[i];
+      const arrayElement = part === undefined
+        || part === null
+        ? ''
+        : part;
```
tiserett: looks kinda complicated, let's think how we can change it

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,29 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let returnString = '';
+    let valueSeparatore = separator;
+
+    if (separator === null) {
+      valueSeparatore = 'null';
+    }
```
tiserett: not sure if you should do it here, I guess if you pass null separator will take it's default value

tiserett (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let returnString = '';
+    const valueSeparatore = separator;
```
tiserett: what is this variable for?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      const part = this[i];
```
tiserett: part of what?
I guess 'item' or 'element' are more used names in such cases

tiserett (changes requested):

src/arrayMethodJoin.js
vserhiichuk: I thought it was considered bad practice to use function parameters directly

tiserett (approved):

PR #2677, opened 2023-08-03
task solved

TommasoGiacopini:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+    let joinedString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        joinedString += this[i];
+      }
+
+      if (i !== this.length - 1) {
+        joinedString += separator;
+      }
```
maxim2310, line 17: you can not doing this condition check every iteration, and just after loop remove last element from string

maxim2310 (approved):
Nice work

PR #2678, opened 2023-08-03
initial commit

maksym-mishchanchuk:


AndrMar1939 (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringAfterJoin = '';
+    const copyArray = this.slice();
```
dffUqp: 2. [CODE STYLE] - avoid creating needless copies of arrays, you can iterate existing array

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+    let finallySeparator = separator;
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    if (finallySeparator === null) {
+      finallySeparator = 'null';
+    }
```
dffUqp: little advice for you

```js
console.log(String(null)) // 'null'
console.log(typeof String(null)) // string

const test = null 

console.log(`${test}`) // 'null'
console.log("result: " + test) // result: null
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,32 @@
...
+
+    for (let i = 0; i < copyArray.length; i++) {
+      if (copyArray[i] === undefined || copyArray[i] === null) {
```
dffUqp: try to invert this condition i think if we go in this direction it might help us to write shorter code
```suggestion
      if (copyArray[i] != null) {
```

dffUqp (changes requested):
Almost done:smiley:

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringAfterJoin = '';
+    const finallySeparator = separator;
```
Moroz-Dmytro: Why do we need this?

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+
+    if (this.length === 0) {
+      return '';
+    }
```
Moroz-Dmytro: Do we really need this?
If length === 0 we will skip the loop and return `stringAfterJoin` which is equal to ''

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,25 @@
...
+        stringAfterJoin += this[i];
+      }
+
+      if (i < this.length - 1 && finallySeparator !== undefined) {
```
Moroz-Dmytro: You set a default value for the separator, so it will never be equal to `undefined`

Moroz-Dmytro (changes requested):
Let's make your code cleaner

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+    let stringAfterJoin = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const element = this[i];
+
+      if (element !== null && element !== undefined) {
+        stringAfterJoin += element;
+      }
+
+      if (i < this.length - 1) {
```

tiserett (approved):

PR #2679, opened 2023-08-03
Solution

Orpheus29:


maksym-mishchanchuk (approved):
Like it!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringFromJoinedArray = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const itemContent
+        = this[i] === null || this[i] === undefined
+          ? ''
+          : this[i];
```
tiserett: I suggest using 'if' block here

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+
+      stringFromJoinedArray
+        += i === 0
+          ? itemContent
+          : `${separator}` + itemContent;
```
tiserett: same here

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringFromJoinedArray = '';
```
tiserett: I guess we can just name it as a 'string'

tiserett (changes requested):

src/arrayMethodJoin.js
Orpheus29: Is it for the sake of readability or why?

Moroz-Dmytro (approved):
Well done 👍

PR #2680, opened 2023-08-04
solution

Yevheniia-Sid:


desevoker (approved):
Good job! 🚀

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringResult = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if ((this[i] !== undefined) && (this[i] !== null)) {
+        stringResult += this[i];
+      }
```

tiserett (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+    let stringResult = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const arrayItem = this[i];
+
+      if ((arrayItem !== undefined) && (arrayItem !== null)) {
+        stringResult += arrayItem;
+      }
+
+      if (i !== this.length - 1) {
```

tiserett (approved):

PR #2681, opened 2023-08-04
add task solution

yar0man:


Orpheus29 (approved):
Very clean and simple one, I like it)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        resultString += this[i];
+      }
```

tiserett (changes requested):

tiserett (approved):
good job, I also realized that you can create variable for this.length, cause you use it twice

PR #2682, opened 2023-08-04
solution

AndrMar1939:


ydratsevska (approved):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const currentArrayLength = this.length;
+    const joinSeparator = separator + '';
```
dffUqp: I propose to use a more readable string transform for it.

```suggestion
    const joinSeparator = String(separator);
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+    let resultingString = '';
+
+    if (!currentArrayLength) {
+      return resultingString;
+    }
+
+    this.forEach((item, index) => {
+      /* const stringCharacter = item ?? '' is the best choice,
+      but I couldn`t disable eslint for the nullish operator */
+      const stringCharacter = item === null || item === undefined
```
dffUqp: You can make it shorter using `==`
```js
console.log(undefined == null) // true

console.log(null == null)  // true

console.log(false == null) // false
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+        ? ''
+        : item;
+
+      currentArrayLength - 1 === index
+        ? resultingString += stringCharacter
+        : resultingString += stringCharacter + joinSeparator;
```
dffUqp: I think better to use if statement here because it makes our code more readable.

dffUqp (changes requested):
Almost done:smiley:

src/arrayMethodJoin.js
AndrMar1939: using == is very dangerous for our karma)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const currentArrayLength = this.length;
+    const joinSeparator = String(separator);
```
Moroz-Dmytro: Do we really need this?
I guess result will be the same with and without `String()`
image

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+    let resultingString = '';
+
+    if (!currentArrayLength) {
+      return resultingString;
+    }
```
Moroz-Dmytro: Do we really need this?
If length === 0 we will skip `forEach` and also return `resultingString`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,30 @@
...
+
+      if (item === null || item === undefined) {
+        stringCharacter = '';
+      }
+
+      if (currentArrayLength - 1 === index) {
+        resultingString += stringCharacter;
+      } else {
+        resultingString += stringCharacter + joinSeparator;
+      }
```
Moroz-Dmytro: Do we really need `else` here? Maybe we can simplify this and use only `if`?
P.S. We do the same part of logic in `if` and in `else` block

Moroz-Dmytro (changes requested):
GJ 🚀 
But let's make your code clearer

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const joinSeparator = String(separator);
+    let resultingString = '';
+
+    this.forEach((item, index) => {
```
tiserett: Using Array methods is prohibited in this task. Use loops, index access, and array length if needed.

src/arrayMethodJoin.js
tiserett: @AndrMar1939 reply to it

tiserett (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,27 @@
...
+      }
+
+      if (this.length - 1 === i) {
+        resultingString += stringCharacter;
+
+        break;
+      }
+
+      resultingString += stringCharacter;
+      resultingString += separator;
```
tiserett, line 24: let's simplify this

tiserett (approved):

PR #2683, opened 2023-08-04
Solution

desevoker:


viktoria-roik (approved):
Good Job!

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,27 @@
...
+        string += separator;
+      }
+
+      let item = this[i];
+
+      if (item === null || item === undefined) {
+        item = '';
+      }
+
+      string += item;
```
Moroz-Dmytro:
```suggestion
      if (item !== null && item !== undefined) {
        string += item;
      }
```

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,27 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        string += separator;
+      }
+
+      let item = this[i];
```
Moroz-Dmytro:
```suggestion
      const item = this[i];
```

Moroz-Dmytro (approved):
LGTM 🔥 
But a small piece of advice, as for me it will be clearer

PR #2687, opened 2023-08-04
add task solution

OlhaMomot:


polinavafik (commented):
great job!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
+
+    if (this.length === 0) {
+      return '';
+    }
```
00Mass00: You don't need this check if you have this one:
```

    if (!this[this.length - 1]) {
      return resultString;
    }

**Review by 00Mass00 — APPROVED** [review]
LGTM!

File: src/arrayMethodJoin.js [hunk 2]
```diff
@@ -4,8 +4,26 @@
...
+
+    for (let i = 0; i < this.length - 1; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
```

Moroz-Dmytro (changes requested):
Try to refactor your code
There is a bug 🐞 with current solution
If I call your method on next array

image

I got next result 

image

But next was expected:

image

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,26 @@
...
+      resultString += separator;
+    }
+
+    const lastElement = this[this.length - 1];
+
+    if (lastElement === undefined || lastElement === null) {
+      return resultString;
+    }
+
+    return resultString + this[this.length - 1];
```
tiserett, line 26: acn we use lastElement here?

tiserett (approved):

PR #2688, opened 2023-08-04
Solution

neovendano:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let arrayString = '';
+    const separatorString = String(separator);
```
polinavafik: separator always will be a string, so you can remove line 9, and on line 17 change separatorString to separator
But if it works, it works! good job🤝👍

polinavafik (approved):

src/arrayMethodJoin.js
dffUqp: It's good comment, remove it 

[5-ways-to-convert-a-value-to-string-in-javascript](https://medium.com/dailyjs/5-ways-to-convert-a-value-to-string-in-javascript-6b334b2fc778)

dffUqp (changes requested):
Almost done:smiley:

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let arrayString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        arrayString += String(this[i]);
+      }
```

tiserett (changes requested):

src/arrayMethodJoin.js
neovendano: Ok

dffUqp (approved):
LGTM🚀

PR #2691, opened 2023-08-05
solution

00Mass00:


yuliiaant (approved):
Good job!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let customJoin = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        customJoin += this[i];
+      }
```

tiserett (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let customJoin = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const index = this[i];
```
tiserett, line 11: better name it as 'item' or 'element'

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+
+      if (index !== null && index !== undefined) {
+        customJoin += index;
+      }
+
+      if (i !== this.length - 1) {
```
tiserett, line 17: I also suggest putting this.length in a variable, since we already use it twice, but I guess it's optional

tiserett (approved):

PR #2692, opened 2023-08-06
solution

Valentyn-Bilyk:


rostyslav48 (approved):
Good Job👏. Everything seem to be fine😎

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
+
+    for (const element of this) {
+      if (element !== null && element !== undefined) {
+        resultString += `${element}`;
```
tiserett, line 12: I guess it's better to use String(), to set something as a string

tiserett (approved):

PR #2693, opened 2023-08-06
Solution

serhiilapin:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,23 @@
...
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function (separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        result += `${this[i]}`;
```
PolinaRomanenko, line 14: You're using template literals for a single variable. It's more efficient to use the variable directly without template literals. Replace `${this[i]}` with this[i].

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,23 @@
...
+      }
+    }
+
+    return result;
   };
 }
 
-module.exports = applyCustomJoin;
+module.exports = applyCustomJoin;
\ No newline at end of file
```
PolinaRomanenko, line 26: Files should end with a newline to comply with the POSIX standard. This ensures better compatibility across different systems and tools.

PolinaRomanenko (approved):
look at the comments

PR #2694, opened 2023-08-06
add task solution

cty2802:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,20 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let joined = '';
+    const formattedSeparator = separator === undefined ? ',' : '' + separator;
```
Vayts: I think Default parameters would be better. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

Vayts (approved):
good job!

tiserett (approved):

PR #2696, opened 2023-08-06
create arr.join method

rostyslav48:


Oleh0293 (approved):
why not just return
joinedStr += this[i];
joinedStr += separator;
In general, the solution is correct, but I would replace these lines.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedStr = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        joinedStr += `${this[i]}`;
```
Moroz-Dmytro:
```suggestion
        joinedStr += this[i];
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+      }
+
+      if (i + 1 === this.length) {
+        break;
+      }
+
+      joinedStr += `${separator}`;
```
Moroz-Dmytro:
```suggestion
      if (i < this.length - 1) {
        joinedStr += separator;
      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedStr = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
```
Moroz-Dmytro: Also, you are using `this[i]` 3 times. Create variable for this value

Moroz-Dmytro (changes requested):
GJ 🚀 
But let's improve your code

src/arrayMethodJoin.js
rostyslav48: I am using `` this syntax in order to prevent errors if there are some numbers in an array

src/arrayMethodJoin.js
rostyslav48: I followed the rule "not write the main code in IF "

Moroz-Dmytro: Here we have only 1 line of code in `if` block, so it's ok
In my opinion, my suggestion is easier for reading and understanding. If you want you can leave it as it is, it's not critical, it's just a recommendation to make your code a little bit easier for understanding

src/arrayMethodJoin.js
Moroz-Dmytro: What happens if we will add `string` to `number`, what type result will have?
For example 'Some text'+4=?

rostyslav48: Oh, yep it works, I was a little bit confused by another programming language which I learn previously. But actually, is it a good practice to add a number to a string?

src/arrayMethodJoin.js
rostyslav48: Ok, thank you!

Moroz-Dmytro (approved):

PR #2699, opened 2023-08-07
Solution

Oleh0293:


Taras2705 (commented):
good job !!!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+    const LENGTH_OF_ARRAY = this.length;
```
voronine: I'm not entirely sure, but it's not easier to immediately use this.length, instead of a separate variable LENGTH_OF_ARRAY.

voronine (approved):
Good job,  all ok.

src/arrayMethodJoin.js
Moroz-Dmytro: If you want to create variable for this, name it in camelCase

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+
+    for (let i = 0; i < LENGTH_OF_ARRAY; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
```

Moroz-Dmytro (changes requested):
Good work ⭐ 
But small comments

tiserett (approved):

PR #2702, opened 2023-08-07
solution arrayMethodJoin

ihorid:


annak1104 (approved):
Great job!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+      switch (this[i]) {
+        case null:
+        case undefined:
+          joinedString += `${separator}`;
+          break;
+        default:
+          joinedString += `${separator}` + `${this[i]}`;
+          break;
+      }
+    }
```
tiserett: I suggest getting rid of switch, let's use if block instead

tiserett (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+        joinedString += `${this[i]}`;
+      }
+
+      if (i < this.length - 1) {
```
tiserett: I suggest putting this.length in a variable, since we already use it twice, but I guess it's optional

tiserett (changes requested):

roman-mirzoian (approved):
Well done 👍

PR #2703, opened 2023-08-07
add task solution

AngerDESTROYS:


Vitalii-Petruk (approved):
Good job! 👍 
Looks great, but you need to think better about the name variable "result". Maybe you should pick a better one...

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
```
Moroz-Dmytro: You are using `this[i]` 3 times. Crate variable for this value

Moroz-Dmytro (changes requested):
GJ 🚀 
Small comment

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
...
+  [].__proto__.join2 = function(SEPARATOR = ',') {
     // write code here
+    let RESULT = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const CURRENT_NUMBER = this[i];
+
+      if (CURRENT_NUMBER !== null && CURRENT_NUMBER !== undefined) {
+        RESULT += CURRENT_NUMBER;
+      }
```
alexander-ignatow: Are you sure this is correct?

Please read method join spec and check out how it handles sparse arrays, null/undefined values

(ex. see here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join)

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(SEPARATOR = ',') {
     // write code here
+    let RESULT = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const CURRENT_NUMBER = this[i];
```
alexander-ignatow: the current number is not the perfect name here - after all you can have an array of anything, not only numbers

In addition, see my comment on naming at the top.

alexander-ignatow (commented):
Good, but some improvements are needed

One general thing - please use variable naming correctly:

For function arguments, function names, and most variables, you need to use `camelCaseNaming`
`UPPER_SNAKE_CASE` is used for constants
`PascalCaseNaming` is used for classes and react components

StasSokolov1 (changes requested):
Hi, look please comments above from Alexander, leave to fix only few things)

StasSokolov1 (changes requested):
Good job, small comments

PR #2704, opened 2023-08-07
solution

Taras2705:


AngerDESTROYS (approved):
Nice!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let string = '';
+    const length = this.length;
+
+    for (let i = 0; i < length; i++) {
+      if (i > 0 && separator !== undefined) {
```
Moroz-Dmytro: You don't need to check if `separator !== undefined`, as you set a default value for `separator`, it will never be equal to `undefined`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,21 @@
...
+        string += separator;
+      }
+
+      if (this[i] !== undefined && this[i] !== null) {
```
Moroz-Dmytro: You are using `this[i]` 3 times. Crate variable for this value

Moroz-Dmytro (changes requested):
GJ 🚀 
But small comments

Moroz-Dmytro (approved):

PR #2707, opened 2023-08-07
Solution

Vitalii-Petruk:


k1NAs (approved):
nicely done! :muscle:

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let separatedArray = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        separatedArray += this[i];
+      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let separatedArray = '';
```
tiserett: you're actually joining values of array, so this value is not really good in our case

tiserett (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedValuesFromArray = '';
+
+    for (let index = 0; index < this.length; index++) {
+      if (this[index] !== null && this[index] !== undefined) {
+        joinedValuesFromArray += this[index];
+      }
```
tiserett: I didn't mean you to replace 'i' with 'index'
Developers always use 'i' as variable for iteration

I wanted you to create a variable such as ```const element = this[i] ```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+
+      if (index < this.length - 1) {
```
tiserett: I suggest putting this.length in a variable, since we already use it twice, but I guess it's optional

tiserett (changes requested):

src/arrayMethodJoin.js
Vitalii-Petruk: roger that

dffUqp (approved):
Looks good🔥

PR #2708, opened 2023-08-07
Solution

Val-Bryks:


src/arrayMethodJoin.js
```diff
@@ -5,8 +5,39 @@
...
+
+    if (separator === undefined) {
+      newSeparator = ',';
+    }
+
+    if (separator === null) {
+      newSeparator = 'null';
+    }
+
+    for (let i = 0; i < this.length - 1; i++) {
```
anastasiiavorobiova:
```suggestion
    for (let i = 0; i < this.length; i++) {
```
Consider looping for every element and check in the loop: if the index is the last element - don't add the separator

src/arrayMethodJoin.js
```diff
@@ -5,8 +5,39 @@
...
+      if (this[i] === null || this[i] === undefined) {
+        result += newSeparator;
+      } else {
+        result += this[i] + newSeparator;
+      }
+    }
+
+    if (this[this.length - 1] === undefined) {
```
anastasiiavorobiova: This additional check is redundant, consider removing it

anastasiiavorobiova (changes requested):
Great work! Added tiny comments to fix

anastasiiavorobiova (approved):
Great work!

PR #2709, opened 2023-08-07
Solution

voronine:


AnastasiiaHombalevska (approved):
good job!

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,26 @@
...
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let stringJoin = '';
+
+    if (this.length === 0) {
+      return stringJoin;
+    }
```
tiserett: not a single iteration will work, if length is 0
you will return stringJoin anyway, so this code is redundant

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,26 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        stringJoin += `${this[i]}`;
+      }
+
+      if (i !== this.length - 1) {
+        stringJoin += separator;
+      }
+    }
```

tiserett (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,24 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
+  [].__proto__.join2 = function(separator = ',') {
     // write code here
+    let stringJoin = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const THIS_ITEM = this[i];
```
tiserett, line 12: it's not type of variable which is used in many places around your code (it's not global), it's just a part of this function, so it shouldn't be written in uppercase, just camelCase

tiserett (approved):

PR #2711, opened 2023-08-07
add task solution

edgergard:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] === null || this[i] === undefined) {
+        joinString += '';
+      } else {
+        joinString += this[i];
+      }
```
tiserett: let's think how we can get rid of else block + create variable for this[i]

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+
+      if (i !== this.length - 1) {
```
tiserett: I also suggest putting this.length in a variable. since you're using it twice already, but I guess it's optional

tiserett (changes requested):

tiserett (approved):

PR #2712, opened 2023-08-07
add task solution

k1NAs:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,19 @@
...
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        string += this[i];
+      }
+
+      if (i < this.length - 1) {
```
tiserett, line 18: I also suggest putting this.length in a variable, since we already use it twice, but I guess it's optional

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,19 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        string += this[i];
+      }
```

tiserett (changes requested):

src/arrayMethodJoin.js
BudnikOleksii: You created a variable but forgot to use it here.

BudnikOleksii (approved):
Good job!
Don't forget to use the variables that are declared.

PR #2713, opened 2023-08-07
Solution

AnastasiiaHombalevska:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let modifyArray = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        modifyArray += `${this[i]}`;
+      }
```

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
+
+      if (i !== this.length - 1) {
```
tiserett: I also suggest putting this.length in a variable, since we already use it twice, but I guess it's optional

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let modifyArray = '';
```
tiserett: modifyArray sounds like an action, let's rename it to modifiedArray

tiserett (changes requested):
I accidentaly selected comment, instead of reject in the previous one ;)

BudnikOleksii (changes requested):
Please, check the comments of the previous interviewer

AnastasiiaHombalevska:
> 

I have not yet worked on making changes or re-qualified.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let modifiedArray = '';
+    const lengthOfElements = this.length;
```
roman-mirzoian, line 9: `this` - is a reference to the original array we're calling this function on, so this variable could be called `arrayLength`.

roman-mirzoian (approved):
Well done 👍

PR #2714, opened 2023-08-07
Develop

bogemna:


serkrops (approved):
looks good. 
but I think there are too many comments...

serkrops (approved):

KlymSudakov (changes requested):
Looks good. Delete all coments and it's 'string', not 'strign' :)

ghost (changes requested):
Not bad!
But you shouldn`t leave so many comments!
Delete them and we can move on

00Mass00 (approved):
Good job!

PR #2715, opened 2023-08-08
Develop

vladyslav-svorin:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const arrCopy = [ ...this ];
+    let ConcatenatedStr = '';
```
VitaliyBodnar: variables name should be in a camelCase

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+
+    if (arrCopy.length === 0) {
```
VitaliyBodnar:
```suggestion
    if (!arrCopy.length) {
```

VitaliyBodnar (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+      if (separator === null) {
+        arrCopy[i] += 'null';
+      }
+
+      concatenatedStr += arrCopy[i];
+    }
+
+    concatenatedStr += arrCopy[arrCopy.length - 1];
+
+    return concatenatedStr;
```
etojeDenys: we can simplify this logic using only one loop and without copy the array
think about how to implement it and feel free to ask in the chat

etojeDenys (changes requested):

loralevitska (approved):

PR #2716, opened 2023-08-08
finish solution

serkrops:


mr-olezha (approved):
good job

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+    const mainArrayClone = [];
+    let mainSeparatorClone = separator;
+
+    if (this.length === 0) {
+      return '';
+    }
+
+    if (separator === null) {
+      mainSeparatorClone = 'null';
+    }
```
tiserett: why would you need this, I guess if there's null it will just be ','

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    const mainArrayClone = [];
+    let mainSeparatorClone = separator;
+
+    if (this.length === 0) {
+      return '';
+    }
```
tiserett: if length === 0 there won't be any iterations so string should be empty anyway

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,33 @@
...
+      } else {
+        mainArrayClone[i] = this[i];
+      }
+    }
+
+    let resultString = `${mainArrayClone[0]}`;
+
+    for (let i = 1; i < mainArrayClone.length; i++) {
+      resultString += mainSeparatorClone + mainArrayClone[i];
+    }
```
tiserett: you can get rid of array and use only one for loop, let's simplify this

tiserett (changes requested):

src/arrayMethodJoin.js
serkrops: I need it for this test:
null_test

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
+      if ((this[i] === null || this[i] === undefined)
+      && i !== this.length - 1) {
+        resultString += '' + mainSeparatorClone;
+      } else if (this[i] === null || this[i] === undefined) {
+        resultString += '';
+      } else if (i === this.length - 1) {
+        resultString += this[i];
+      } else {
+        resultString += this[i] + mainSeparatorClone;
+      }
```
dffUqp: This logic looks too complicated, let's rewrite it using 2 `if` statements. 

To achieve it we need to check: 
1. if our value is not equal to `null` and `undefined` 
2. if it is not a last iteration



Small tip - create a separate variable for `this[i]`, like `currentElem`

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,28 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
+    let mainSeparatorClone = separator;
+
+    if (separator === null) {
+      mainSeparatorClone = 'null';
+    }
```
dffUqp: Try to use this reference to remove this `if` statement



```js
console.log(String(null)) // 'null'
console.log(typeof String(null)) // string

const test = null 

console.log(`${test}`) // 'null'
console.log("result: " + test) // result: null
```

Feel free to write me in slack if you have any problems

[5-ways-to-convert-a-value-to-string-in-javascript](https://medium.com/dailyjs/5-ways-to-convert-a-value-to-string-in-javascript-6b334b2fc778)

dffUqp (changes requested):
Almost done:smiley:

33Orange (approved):
Well done

PR #2718, opened 2023-08-08
Solution

okanivan:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        result += this[i];
+      }
```

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,20 @@
...
+
+      if (i < this.length - 1) {
```

tiserett (changes requested):

src/arrayMethodJoin.js
```diff
@@ -4,9 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let result = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const item = this[i];
```
roman-mirzoian, line 11:
```suggestion
      const currentItem = this[i];
```

roman-mirzoian (approved):
Well done 👍

PR #2719, opened 2023-08-08
add solution for the task

thaiduk:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+    let string = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const stringPart = this[i];
+
+      if (stringPart !== null && stringPart !== undefined) {
+        string += stringPart;
+      }
+
+      if (i < this.length - 1) {
```

tiserett (approved):

PR #2722, opened 2023-08-09
Solution

KlymSudakov:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+    for (let i = 0; i < this.length; i++) {
+      if (i > 0) {
+        joinResult += separator;
+      }
+
+      if (this[i] === null || this[i] === undefined) {
+        continue;
+      }
+
+      joinResult += this[i];
```
roman-mirzoian, line 19: You're calling `this[i]` three times, so it would be cool to create a variable for that and reuse it.

roman-mirzoian (approved):
Well done, but pay attention to the comment.

PR #2723, opened 2023-08-09
solution

mykyta01:


src/arrayMethodJoin.js
```diff
@@ -6,6 +6,25 @@
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
mykhalenych, line 7: yoy can add default value for argument

src/arrayMethodJoin.js
```diff
@@ -6,6 +6,25 @@
...
     // write code here
+    let delimiter = separator;
```
mykhalenych, line 9: and delete this logic

mykhalenych (approved):

PR #2732, opened 2023-08-13
Solution

Nikitavyskreb:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let stringArray = '';
```
roman-mirzoian, line 8: Why do you use the word `array` in the title, but it is a string type?
Try to choose a more meaningful naming.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
...
+
+    for (let i = 0; i < this.length; i++) {
+      const arrayElement = this[i];
```
roman-mirzoian, line 11: Try to choose a more meaningful naming.

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.
Correct naming is very important.

PR #2736, opened 2023-08-15
Solution

patrykyo20:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    let result = '';
+
+    if (separator === undefined) {
+      // eslint-disable-next-line no-param-reassign
+      separator = ',';
+    }
```
VitaliyBodnar, line 13: you don't need to mutate argument of function. You can just set default value for the `separator `
```suggestion
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
VitaliyBodnar, line 7:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

VitaliyBodnar (changes requested):

PR #2741, opened 2023-08-16
solution

oleg-matvienko:


src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
...
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== null && this[i] !== undefined) {
+        resultString += this[i];
+      }
```
roman-mirzoian: We use `this[i]` three times, so we can put it in a variable.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,20 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let resultString = '';
```
roman-mirzoian: The variable name is fine, but I would add a more precise name.
```suggestion
    let joinedString = '';
```

roman-mirzoian (changes requested):
Good job, but there is quite a bit of work to be done.

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,23 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+    let value = [];
```
roman-mirzoian: It is not necessary to bring the definition of a variable that is used in the middle of the loop outside.
You should also think about better naming.

roman-mirzoian (changes requested):
Almost done!

src/arrayMethodJoin.js
```diff
@@ -4,8 +4,22 @@
  * Implement method join
  */
 function applyCustomJoin() {
-  [].__proto__.join2 = function(separator) {
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let joinedString = '';
+
+    for (let i = 0; i < this.length; i++) {
+      const valueOfArray = this[i];
```
Moroz-Dmytro, line 11: But still not a good name

Moroz-Dmytro (approved):
👍

PR #2742, opened 2023-08-20
complete the tusk

Annbpiu:
task*)

### DIFF
--- a/src/arrayMethodJoin.js
+++ b/src/arrayMethodJoin.js
@@ -5,7 +5,29 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    if (this.length === 0) {
+      return '';
+    }
+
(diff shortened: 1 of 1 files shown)


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,29 @@
...
-    // write code here
+    if (this.length === 0) {
+      return '';
+    }
+
+    let finalSeparator = separator;
+
+    if (separator === undefined) {
+      finalSeparator = ',';
+    }
```
```suggestion
```
looks like redundant.
instead to check separator on undefined set `','` as default argument.

VitaliyBondarenko1982 (changes requested):
Let's simplify solution

VitaliyBondarenko1982 (changes requested):
.

VitaliyBondarenko1982 (approved):
GJ!

PR #2743, opened 2023-08-21
stage-1

no4kar:


src/arrayMethodJoin.js
```diff
@@ -1,11 +1,51 @@
 'use strict';
 
-/**
- * Implement method join
- */
 function applyCustomJoin() {
+  /**
+   * @param {string | undefined} separator
+   * @returns {string} */
```
denys-danyliuk, line 6: you don't need this

src/arrayMethodJoin.js
```diff
@@ -1,11 +1,51 @@
...
   [].__proto__.join2 = function(separator) {
```
denys-danyliuk, line 7: add default value to the separator in case it wasn't passed

denys-danyliuk (approved):

denys-danyliuk (approved):

PR #2744, opened 2023-08-21
Solution added

kyrylomanko:


src/arrayMethodJoin.js
```diff
@@ -4,9 +4,20 @@
...
-    // write code here
+  [].__proto__.join2 = function(separator = ',') {
+    let str = '';
+
+    for (let i = 0; i < this.length; i++) {
+      if (this[i] !== undefined && this[i] !== null) {
+        str += this[i];
+      }
+
+      if (i !== this.length - 1) {
```
lerastarynets, line 15: Instead of checking if the current iteration is not the last, you can simplify your code by adding the separator after each iteration, and then remove the last separator at the end.

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #2748, opened 2023-08-25
added solution task

shikoryak:


src/arrayMethodJoin.js
```diff
@@ -5,7 +5,24 @@
...
+
+      if (value === undefined || value === null) {
+        value = '';
+      }
+
+      if (i === 0) {
+        result += value;
+      } else {
+        result += readySeparator + value;
+      }
```
```suggestion
      let value = this[i];

      if (value !== undefined && value !== null) {
        result += value;
      }

      if (i !== this.length - 1) {
        result += readySeparator;
      } 
```

etojeDenys (changes requested):

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,22 @@
  */
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
```
IvanFesenko:
```suggestion
  [].__proto__.join2 = function(separator = ',') {
```

src/arrayMethodJoin.js
```diff
@@ -5,7 +5,22 @@
...
-    // write code here
+    let result = '';
+    const readySeparator = separator !== undefined ? String(separator) : ',';
```
IvanFesenko, line 9: default value looks more readable

IvanFesenko (approved):
Well done

IvanFesenko (approved):
.

IvanFesenko (approved):
One more try

PR #2751, opened 2023-09-15
add task solution

andriy-chornyy:


src/arrayMethodJoin.js
```diff
@@ -3,9 +3,30 @@
 /**
  * Implement method join
  */
+
 function applyCustomJoin() {
   [].__proto__.join2 = function(separator) {
-    // write code here
+    const { sep } = { sep: separator === undefined ? ',' : String(separator) };
```
anastasiiavorobiova, line 9: `const sep = separator === undefined ? ',' : String(separator);`

src/arrayMethodJoin.js
```diff
@@ -3,9 +3,30 @@
...
+      if (i > 0) {
+        result += sep;
+      }
+
+      if (this[i] === null || this[i] === undefined) {
+        result += '';
+      } else {
+        if (String(this[i]) === 'null') {
+          result += 'null';
+        } else {
```
anastasiiavorobiova, line 23:
```
if (String(this[i]) === 'null') {
          result += 'null';
        } else {

```
You can delete these lines of code

anastasiiavorobiova (changes requested):
Great work! Almost done!

